PR #3, opened 2013-10-12
Fixing the form

Berdir:


agentrickard:
Found the same solution at the same time.

PR #7, opened 2014-09-30
@see https://www.drupal.org/node/2343661

franzwilding:


agentrickard:
Actually _l is wrong as well. I've fixed this in most places.

PR #32, opened 2014-10-27
30 setters

agentrickard:


siilak:
Testing it. What lines I should add to configuring settings.php file?
In drupal 7, I had:
include DRUPAL_ROOT . '/path/to/modules/domain/settings.inc';

agentrickard:
None. We don't need that in D8.

siilak:
Ok, thanks :)

PR #36, opened 2015-02-17
Fix domain id references

franzwilding:
I think this will work. The domain_id was not saved to the database, because in DomainForm the wrong id was accessed. 

domain_access_node_grants also needed some updates, so we can get the domain_id and not the machinename.

### DIFF
--- a/domain_access/domain_access.module
+++ b/domain_access/domain_access.module
@@ -38,15 +38,13 @@ function domain_access_node_grants(AccountProxy $account, $op) {
   }
 
   // Grants for edit/delete require permissions.
-  if($account->isAuthenticated()) {
-    $user = entity_load('user', $account->id());
-    $user_domains = domain_access_get_entity_values($user, DOMAIN_ACCESS_USER_FIELD);
-    if ($op == 'update' && $account->hasPermission('edit domain content') && isset($user_domains[$id])) {
-      $grants['domain_id'][] = $id;
-    }
-    if ($op == 'delete' && $account->hasPermission('delete domain content') && isset($user_domains[$id])) {
-      $grants['domain_id'][] = $id;
-    }
+  $user = entity_load('user', $account->id());
+  $user_domains = domain_access_get_entity_values($user, DOMAIN_ACCESS_USER_FIELD);
+  if ($op == 'update' && $account->hasPermission('edit domain content') && isset($user_domains[$id])) {
+    $grants['domain_id'][] = $id;
+  }
+  if ($op == 'delete' && $account->hasPermission('delete domain content') && isset($user_domains[$id])) {
+    $grants['domain_id'][] = $id;
   }
 
   return $grants;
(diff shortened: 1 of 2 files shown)


domain_access/domain_access.module
```diff
@@ -25,22 +26,27 @@ define('DOMAIN_ACCESS_USER_FIELD', 'field_domain_user');
...
   }
+
   // Grants for edit/delete require permissions.
-  $user_domains = domain_access_get_entity_values($account, DOMAIN_ACCESS_USER_FIELD);
-  if ($op == 'update' && $account->hasPermission('edit domain content') && isset($account_domains[$id])) {
-    $grants['domain_id'][] = $id;
-  }
-  if ($op == 'delete' && $account->hasPermission('delete domain content') && isset($account_domains[$id])) {
-    $grants['domain_id'][] = $id;
+  if($account->isAuthenticated()) {
```
agentrickard: missing space

agentrickard: It is theoretically possible that someone would give this permission to an anonymous user. Imagine a scenario where there are two sites: notes.examplex.com and published.example.com. Anon users might be able to post / edit / delete on notes.example.com, with best content moving to published.example.com.

It's an odd edge-case, but I don't think we should be restrictive here by forcing a check for authentication.

franzwilding:
I think you are right, but we need to get the domains, assigned to an user in domain_access_node_access_records, so wie need an user object, aren't we?

agentrickard:
Yes, but can't you run an entity_load on UID 0?

I think in earlier iterations of the code, we assume that the $account object is already loaded.

franzwilding:
You are absolutely right.

domain_access/domain_access.module
```diff
@@ -5,12 +5,13 @@
  * Domain-based access control for content.
  */
 
-use Drupal\domain\DomainNegotiator;
+use Drupal\domain\DomainManager;
```
agentrickard, line 8: This is outdated. We switched to using a Negotiator class a while back to be consistent with core naming conventions.

domain_access/domain_access.module
```diff
@@ -25,21 +26,24 @@ define('DOMAIN_ACCESS_USER_FIELD', 'field_domain_user');
 /**
  * Implements hook_node_grants().
  */
-function domain_access_node_grants($op, AccountInterface $account) {
+function domain_access_node_grants(AccountProxy $account, $op) {
```
agentrickard, line 29: Looks like we want AccountInterface not AccountProxy.

https://api.drupal.org/api/drupal/core!modules!node!node.api.php/function/hook_node_grants/8

agentrickard:
I'm going to ensure we have proper testing for these changes, but may do so in a separate PR.

Please make sure you are testing against Drupal 8 HEAD and the 8.0.x branch.

agentrickard:
Closing in favor of PR #38

Issue #37, opened 2015-02-22
Better Domain Access tests


agentrickard:
We need better tests of the integration to Node Access.

agentrickard:
Committed in https://github.com/agentrickard/domain/pull/38

PR #39, opened 2015-02-25
Add Condition Plugin

franzwilding:
I added an condition Plugin for domains. If you want to include it into domain module, I can create some tests. Otherwise I would create a separate module for this feature.

### DIFF
--- a/domain/src/Plugin/Condition/Domain.php
+++ b/domain/src/Plugin/Condition/Domain.php
@@ -1,0 +1,90 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\domain\Plugin\Condition\Domain.
+ */
+
+namespace Drupal\domain\Plugin\Condition;
+
+use Drupal\Core\Condition\ConditionPluginBase;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Plugin\Context\ContextDefinition;
+
+/**
+ * Provides a 'Domain' condition.
+ *
+ * @Condition(
+ *   id = "domain",
+ *   label = @Translation("Domain")
+ * )
+ *
+ */
+class Domain extends ConditionPluginBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+
+    $options = array();
+
+    foreach (domain_load_and_sort() as $key => $value) {
+      $options[$key] = $value->get('name');
+    }
+
+    $form['domains'] = array(
+      '#type' => 'checkboxes',
+      '#title' => $this->t('Domains'),
+      '#options' => $options,
+      '#default_value' => $this->configuration['domains'],
+      '#description' => $this->t('Select domains for this block to be shown. If no are selected, this block will be visible on all domains.'),
+    );
+
+    return parent::buildConfigurationForm($form, $form_state);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitConfigurationForm(array &$form, FormStateInterface $form_state) {
+    $this->configuration['domains'] = array_filter($form_state->getValue('domains'));
+    parent::submitConfigurationForm($form, $form_state);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function summary() {
+
+    $domains = $this->configuration['domains'];
+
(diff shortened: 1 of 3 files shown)


agentrickard:
SimpleTests are fine, though if we had Unit Tests, that would be cool, too.

domain_access/domain_access.module
```diff
@@ -5,12 +5,13 @@
  * Domain-based access control for content.
  */
 
-use Drupal\domain\DomainNegotiator;
+use Drupal\domain\DomainManager;
```
agentrickard, line 8: This is still written against the wrong branch. Development is done against domain 8.0.x branch, which changed to using DomainNegotiator a long time ago.

domain_access/domain_access.module
```diff
@@ -25,21 +26,24 @@ define('DOMAIN_ACCESS_USER_FIELD', 'field_domain_user');
 /**
  * Implements hook_node_grants().
  */
-function domain_access_node_grants($op, AccountInterface $account) {
+function domain_access_node_grants(AccountProxy $account, $op) {
```
agentrickard, line 29: No. AccountProxy is an implementation of AcoutnInterface. It is proper to type hint with the Interface, because implementations may change.

This has been accounted for over in PR #37

domain/src/Plugin/Condition/Domain.php
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Provides a 'Domain' condition.
+ *
+ * @Condition(
+ *   id = "domain",
+ *   label = @Translation("Domain")
+ * )
+ *
+ */
+class Domain extends ConditionPluginBase {
```
agentrickard, line 23: Is this really the core pattern? To name the condition after the module and not DomainCondition? That seems odd.

domain/src/Plugin/Condition/Domain.php
```diff
@@ -0,0 +1,90 @@
...
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+
+    $options = array();
+
+    foreach (domain_load_and_sort() as $key => $value) {
+      $options[$key] = $value->get('name');
+    }
+
+    $form['domains'] = array(
+      '#type' => 'checkboxes',
```
agentrickard, line 37: From a UI point of view, checkboxes may not scale beyond 20 domains. Are there better options here?

domain/src/Plugin/Condition/Domain.php
```diff
@@ -0,0 +1,90 @@
...
+      '#title' => $this->t('Domains'),
+      '#options' => $options,
+      '#default_value' => $this->configuration['domains'],
+      '#description' => $this->t('Select domains for this block to be shown. If no are selected, this block will be visible on all domains.'),
```
agentrickard, line 41: Block?

domain/src/Plugin/Condition/Domain.php
```diff
@@ -0,0 +1,90 @@
...
+    $domains = $this->configuration['domains'];
+
+    if (count($domains) > 1) {
+      $domains = implode(', ', $domains);
+    }
+    else {
+      $domains = reset($domains);
+    }
+    if (!empty($this->configuration['negate'])) {
+      return $this->t('Visible on all domains extept: @domains', array('@domains' => $domains));
```
agentrickard, line 69: except

domain/src/Plugin/Condition/Domain.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+    else {
+      return $this->t('Visible on this domains: @domains', array('@domains' => $domains));
```
agentrickard, line 72: "this domain" / "these domains"

domain/src/Plugin/Condition/Domain.php
franzwilding, line 23: I think it's the core pattern. I checked the following modules: 
language: Language
node: NodeType
user: UserRole

I also found CurrentThemeCondition in system module, which also looks cleaner to me.

Beside the class name, I'm not quit surre about any context annotations.

agentrickard:
I was confused about the purpose of this, thinking that it was based on our discussions of entity reference.

This should go in the main module, I think. It's easy to add. Let's define the behavior over in https://github.com/agentrickard/domain/issues/42

PR #44, opened 2015-03-20
Adds Condition plugin.

agentrickard:
Odd. This mostly works, but it looks like the summary() and evaluate() methods are not firing.

What are we missing here?


agentrickard:
This appears to be a core bug, at least for summary().

PR #55, opened 2015-07-14
Domain_id datatype error fix

davidDuymelinck:
the id method calls the machine name of the entity which is a string and not an integer, which breaks setting the node_access permissions.

### DIFF
--- a/domain/src/DomainForm.php
+++ b/domain/src/DomainForm.php
@@ -30,7 +30,7 @@ class DomainForm extends EntityForm {
     }
     $form['domain_id'] = array(
       '#type' => 'value',
-      '#value' => $domain->id(),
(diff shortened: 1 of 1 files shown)


agentrickard:
Right. Is this testable?

davidDuymelinck:
When I used the module the domainId for the second domain was null.

agentrickard:
I'm sorry, I mean: can an automated test be written for this?

davidDuymelinck:
You could add following lines to the domainCreateTest

```
// Has domain id been set?
$this->assertTrue($new_domain->getDomainId() == 1, 'Domain id set properly.');
```

agentrickard:
Great. Please add to this PR.

PR #57, opened 2015-07-15
Fix cli node_access_rebuild error

davidDuymelinck:
When you add domains during a drush site install, you need to call node_access_rebuild. This calls domain_access_node_grants, and because the domain_get_domain function is url based you have to create a localhost domain because that is the default. 

To prevent the creation of a domain just because it's the default value, I added an extra check that gets the default domain instead. it doesn't interfere with the permission rebuild via the web interface.

### DIFF
--- a/domain_access/domain_access.module
+++ b/domain_access/domain_access.module
@@ -35,6 +35,10 @@ function domain_access_node_grants(AccountInterface $account, $op) {
     $active = domain_default(); 
   }
   
+  if(empty($active)) {
+    return $grants;
+  }
+  
   $id = $active->getDomainId();
   // Grants for view are simple. Use the active domain.
   if ($op == 'view') {
--- a/domain_access/domain_access.module
+++ b/domain_access/domain_access.module
@@ -31,7 +31,9 @@ function domain_access_node_grants(AccountInterface $account, $op) {
   $grants = array();
   $active = domain_get_domain();
   
-  if(empty($active)){ $active = domain_default(); }
+  if(empty($active)) { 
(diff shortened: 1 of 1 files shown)


agentrickard:
Remember that in the D8 version,. you only have to Node Access rebuild if Domain Access is installed.

domain_access/domain_access.module
```diff
@@ -30,6 +30,9 @@ define('DOMAIN_ACCESS_USER_FIELD', 'field_domain_user');
 function domain_access_node_grants(AccountInterface $account, $op) {
   $grants = array();
   $active = domain_get_domain();
+  
+  if(empty($active)){ $active = domain_default(); }
```
agentrickard: Not formatted to standards. And is it possible that domain_default() is empty?

davidDuymelinck:
Yes Node Access is installed in my case.

domain_access/domain_access.module
davidDuymelinck: If you use node_access_rebuild it will fail. It solves all the scenario's where domains are already set and a node_access_rebuild call is needed.

agentrickard:
Right, the code needs to account for that case. Returning an empty array if no domains are preset should be valid.

davidDuymelinck:
You suggest something like:

```
if(empty($active)) { 
  $active = domain_default(); 
}

if(empty($active)) {
   return $grants;
}
```

agentrickard:
Exactly. Don't let NULL returns throw errors.

PR #77, opened 2015-11-09
Add a composer.json

tstoeckler:
It would be great to be able to install this module via Composer. The really important parts are the "name" and the "type", the rest is nice-to-have. I looked at what some of the other modules are doing and searched the interwebs for your email (hope that was OK), but I don't really care that much about the details, as long as there _is_ a composer.json


agentrickard:
Thanks.

PR #80, opened 2015-11-09
Fix drush example commands

tstoeckler:


domain/domain.drush.inc
```diff
@@ -113,19 +113,19 @@ function domain_drush_command() {
...
-      'name' => 'The name to use for the domain.',
+      'name' => 'The machine-readable name to use for the domain.',
     ),
   );
   $items['domain-scheme'] = array(
     'description' => 'Changes a domain scheme.',
     'examples' => array(
       'drush domain-scheme example.com',
-      'drush domain-scheme 1 Foo',
+      'drush domain-scheme 1',
```
agentrickard, line 128: This should be:

drush domain-scheme 1 http

The arguments descriptions are all wrong too.

agentrickard:
Fixed that up in a followup.

PR #82, opened 2015-11-09
Fix numerous problems with the shipped config

tstoeckler:
I separated the changes into atomic commits, see each of the commit messages for details.


agentrickard:
I wonder why tests didn't catch any of this.

PR #83, opened 2015-11-09
Improve translatability of messages in DomainController

tstoeckler:
I took the liberty of including some minor code style changes in the same file, but as a separate commit in case you want to cherry-pick.


agentrickard:
I just changed these strings, since core deprecated the !argument.

Can you re-roll?

PR #88, opened 2015-11-17
Cache context

bjaxelsen:
I've had some errors making the module work together with the context_caches. I've tried my best but I am not sure this is 100% correct. 

Unfortunately I am not currently able to run tests (there are some translation system issues that blocks my tests) - so I haven't tested the code.


bjaxelsen:
(please ignore the domain_config/src/DomainConfigOverrider.php.orig file - this is my first PR and I made an error)

agentrickard:
You know, I had a Cache Context and then removed it recently, because I think we can use the core `url.site` context in any case where you need a domain-specific cache.

See https://github.com/agentrickard/domain/blob/8.x-1.x/domain_config/src/DomainConfigOverrider.php#L140
And https://github.com/agentrickard/domain/commit/656ba90101354de1c3b78238402cce00d0c59574

Based on my tests with Domain Config, we don't need our own caching system.

domain/domain.services.yml
```diff
@@ -9,6 +9,11 @@ services:
     arguments: ['@domain.negotiator']
     tags:
       - { name: 'context_provider' }
+  cache_context.domain:
+    class: Drupal\domain\Cache\DomainCacheContext
+    arguments: ['@request_stack']
```
agentrickard, line 14: For reference, this would inject the @request_stack into the __consruct() method. That method isn't present, and we don't use the Request Stack, we  use the @domain_negotiator service.

You can see this pattern in https://github.com/agentrickard/domain/blob/8.x-1.x/domain_config/src/DomainConfigOverrider.php#L66

bjaxelsen:
Thanks for the feedback - yes, it does make sense to make url.site do the job. And yes, injecting the @request_stack was a mistake

agentrickard:
No problem. I make mistakes like that all the time.

At my office, we commonly use Pull Requests to provide tips, so that's the pattern I'm following.

Before I close this, what was your use-case?  

Note that by default, I'm using both url.site and language as a cache context in Domain Config, because I think we need to make everything language-aware.

bjaxelsen:
The use case was that I use Domain on two sites and get error messages regarding an invalid  "domain" cache context. I've created a new PR to change it to url.site.

zerolab:
To keep record, the new PR is #89

PR #92, opened 2015-11-26
Fallback to default domain

bjaxelsen:
For convenience, I assume we should fallback to the default domain. This was not currently working when working on a local machine. Can we fix it this way?


agentrickard:
Probably. That's the default behavior in lower versions of the module.

We should also look at this part, which I think I had in just for early testing.

```
    // If a straight load fails, create a base domain for checking.
    if (empty($domain)) {
      $values = array('hostname' => $httpHost);
      $domain = \Drupal::entityManager()->getStorage('domain')->create($values);
    }
```

Creating a domain request on the fly seems like something we should remove.

There is also a typo that needs fixing on line 59 `$loookup`

agentrickard:
I am re-rolling those changes in #94.

PR #110, opened 2016-01-05
Replaced all calls to negotiateActiveDomain with getActiveDomain. Met…

itsekhmistro:
Replaced all calls to negotiateActiveDomain with getActiveDomain.
Method getActiveDomain() uses cached result of negotiateActiveDomain() and is safe for performance.

### DIFF
--- a/domain/src/Access/DomainAccessCheck.php
+++ b/domain/src/Access/DomainAccessCheck.php
@@ -56,7 +56,7 @@ class DomainAccessCheck implements AccessCheckInterface {
    * {@inheritdoc}
    */
   public function access(AccountInterface $account) {
-    $domain = $this->domainNegotiator->negotiateActiveDomain();
+    $domain = $this->domainNegotiator->getActiveDomain();
     // Is the domain allowed?
     // No domain, let it pass.
     if (empty($domain)) {
--- a/domain/src/DomainNegotiatorInterface.php
+++ b/domain/src/DomainNegotiatorInterface.php
@@ -25,11 +25,6 @@ interface DomainNegotiatorInterface {
   public function setActiveDomain(DomainInterface $domain);
 
   /**
-   * Gets the active domain.
-   */
-  public function negotiateActiveDomain();
-
-  /**
    * Stores the inbound httpHost request.
    */
   public function setHttpHost($httpHost);
--- a/domain/src/DomainNegotiatorInterface.php
+++ b/domain/src/DomainNegotiatorInterface.php
@@ -52,6 +52,6 @@ interface DomainNegotiatorInterface {
   /**
    * Gets the active domain.
    */
-  public function getActiveDomain($rest = FALSE);
+  public function getActiveDomain($reset = FALSE);
 
 }
--- a/domain/src/DomainNegotiator.php
+++ b/domain/src/DomainNegotiator.php
@@ -102,9 +102,9 @@ class DomainNegotiator implements DomainNegotiatorInterface {
   }
 
   /**
-   * {@inheritdoc}
+   * Determine the active domain.
    */
-  public function negotiateActiveDomain() {
+  protected function negotiateActiveDomain() {
     $httpHost = $this->negotiateActiveHostname();
     $this->setRequestDomain($httpHost);
     return $this->domain;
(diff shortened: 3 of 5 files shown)


domain/src/Access/DomainAccessCheck.php
```diff
@@ -56,7 +56,7 @@ public function checkPath($path) {
    * {@inheritdoc}
    */
   public function access(AccountInterface $account) {
-    $domain = $this->domainNegotiator->negotiateActiveDomain();
+    $domain = $this->domainNegotiator->getActiveDomain();
```
agentrickard: This has already been fixed. Is your fork up-to-date?

https://github.com/agentrickard/domain/blob/8.x-1.x/domain/src/DomainNegotiator.php#L116

domain/src/DomainNegotiatorInterface.php
```diff
@@ -52,6 +52,6 @@ public function negotiateActiveHostname();
   /**
    * Gets the active domain.
    */
-  public function getActiveDomain($rest = FALSE);
+  public function getActiveDomain($reset = FALSE);
```
agentrickard, line 50: This has already been fixed. Is your fork up-to-date?

https://github.com/agentrickard/domain/blob/8.x-1.x/domain/src/DomainNegotiator.php#L116

agentrickard:
There is a reason why these two methods are separate, and the context of the caller matters. 

negotiateActiveDomain() assumes that this is an internal API call from within the module, generally called from within an early operation -- access check, kernel request -- and should reset the cached state.

getActiveDomain() should be used by external callers to retrieve the results of the negotiation. 

I would only make this conversion if negotiateActiveDomain() were called from a module other than Domain proper.

The real need here is to fix the method documentation in DomainNegotiatorInterface.

agentrickard:
It may be that onKernelRequestDomain() is the only method that should call negotiateActiveDomain().

itsekhmistro:
Hi Ken,

Thank you for feedback.

> It may be that onKernelRequestDomain() is the only method that should call negotiateActiveDomain().

I agree here but it also safe to call getActiveDomain() in the place. Please consider that onKernelRequestDomain() could occur only once on Request according to subscribed events.

> There is a reason why these two methods are separate, and the context of the caller matters.
> - negotiateActiveDomain() assumes that this is an internal API call from within the module, generally called from within an early operation -- access check, kernel request -- and should reset the cached state.
> - getActiveDomain() should be used by external callers to retrieve the results of the negotiation.
>   I would only make this conversion if negotiateActiveDomain() were called from a module other than Domain proper.

During my module development it appeared that domain detection (negotiateActiveDomain()) and related hook_domain_request were called for 3 times. It was called from Domain module itself. This is not good for performance reasons.

As far I see the getActiveDomain() is a safe implementation for getting active domain no matter domain has been already detected or not. It uses the result of negotiation if available, and if not will call for negotiate.

Please let me know what your think about it.

agentrickard:
I thought about this some more and I think we should do the following:
- Make negotiateActiveDomain a private method. That will be an explicit signal to developers not to call it directly.
- Use getActiveDomain() in all external calls.
- In onKernelRequestDomain, make sure to pass $reset = TRUE.

itsekhmistro:
Hi,

1)

> Make negotiateActiveDomain a private method. That will be an explicit signal to developers not to call it directly.

Ok. Private or protected is ok.

2)

> Use getActiveDomain() in all external calls.

Yes, absolutely.

3)

> In onKernelRequestDomain, make sure to pass $reset = TRUE.

Yes, sounds logical. (meanwhile it won't change much - when onKernelRequestDomain is executed the domain is not initialized)

Would you like me to apply (1) and (3) in the pull request?

agentrickard:
Yes, please do.

I'm torn on private v. protected. This class is a service, so can be swapped out entirely. I can't think of a use-case where the class would be extended, but using protected might be smart.

itsekhmistro:
Pull request is updated.

agentrickard:
We also have to change the public/protected flag in the base interface.

```
  /**
   * Gets the active domain.
   */
  public function negotiateActiveDomain();
```

itsekhmistro:
Hi Ken,

> We also have to change the public/protected flag in the base interface.

The interfaces shouldn't contain protected/private methods.

> public function negotiateActiveDomain();

Thus it was removed from the interface declaration. It's already in the diff of pull request.

agentrickard:
Ah, my use of Interfaces is still pretty new, thanks.

itsekhmistro:
Thank you

agentrickard:
We missed one. See #111

Issue #107, opened 2015-12-06
Make language optional for domain overrides


agentrickard:
In the Drupal 8 version, Domain Config is language sensitive (e.g. domain.config.example_com.en.config_item.yml).

We should allow a config setting to be active across an entire domain, with language optional.

That means check first for the domain/language combination, but fall back to domain-only.

* IF domain.config.example_com.en.config_item.yml
* ELSE domain.config.example_com.config_item.yml

agentrickard:
This is causing a circular dependency failure that I don't quite understand.

agentrickard:
This code is sitting in https://github.com/agentrickard/domain/tree/107-optional and it works fine through the UI, but running tests it throws a circular dependency error.

To see, run the Domain Config tests. 

Tests also fail if the 8.x.-1.x test is updated as follows:

```
<?php

/**
 * @file
 * Definition of Drupal\domain_config\Tests\DomainConfigTestBase.
 */

namespace Drupal\domain_config\Tests;

use Drupal\domain\Tests\DomainTestBase;
use Drupal\language\Entity\ConfigurableLanguage;

/**
 * Helper test methods for Domain Config testing.
 */
abstract class DomainConfigTestBase extends DomainTestBase {

  /**
   * Modules to enable.
   *
   * @var array
   */
  public static $modules = array('domain', 'domain_config', 'domain_config_test', 'language', 'locale');

  function setUp() {
    parent::setUp();
    ConfigurableLanguage::createFromLangcode('fr')->save();
  }

}
```

There is something going on inside the LanguageManager that I just can't follow.

agentrickard:
If setting up the test environment is too much, here's the error message.

```
Uncaught PHP Exception Symfony\Component\DependencyInjection\Exception\ServiceCircularReferenceException: "Circular reference detected for service "language_manager", path: "language_request_subscriber -> language_manager -> config.factory -> domain_config.overrider"." at /Applications/MAMP/htdocs/drupal-8/core/lib/Drupal/Component/DependencyInjection/Container.php line 151 
```

mevaibhav:
Hey Rickard, 

Appreciate your efforts migrating this Module to version 8. However , do you have any estimate time to complete this or the first dev version.

Any response would be welcomed

agentrickard:
I do not, since I no longer do development day-to-day.

The APIs are, I believe, stable, so the current state could be considered an alpha.

Issue #145, opened 2016-01-22
Replace / inject the storage manager in DomainAliasLoader.


agentrickard:
See #144

agentrickard:
This turns out to be impractical, given the early request in the bootstrap process.

Issue #144, opened 2016-01-22
Replace / inject the storage manager in DomainLoader.


agentrickard:
This is currently throwing an error when I tried to replace the entity manager, which has been deprecated.

agentrickard:
See https://www.drupal.org/node/2549139

bjaxelsen:
EntityManager has been deprecated, instead we should use EntityTypeManager.

I think it is not possible to inject it through the constructor:

1) When initializing the language manager service (and its dependencies) early during the page request by the Symfony service manager, Domain Config will be activated when loading language configuration.
2) Domain Config will depend on DomainLoader.
2) If we inject EntityTypeManager, it will requires injection of the string translation service.
3) The string translation service requires the language manager service, hence we have a circular dependency.

So we need to avoid this circular dependency and let the language manager service be fully initialized before we depend on it.

I am not quite sure how to solve it, i tried giving it a go in #165 but we may need to double check that this is the right solution.

agentrickard:
That makes sense. I just couldn't follow the logic when the error came up. 

Oddly, I only had trouble during the tests. 

Thanks for tracking this down!

agentrickard:
The fix in Domain Config works, but when I tried to inject the EntityTypeManager into the two loaders, I got the errors during tests again. So I may just drop it for now,

agentrickard:
This is what I tried.

[test.patch.txt](https://github.com/agentrickard/domain/files/110463/test.patch.txt)

agentrickard:
This turns out to be impractical, given the early request in the bootstrap process.

PR #165, opened 2016-01-25
Circular service dependencies and deprecated method

bjaxelsen:
When enabling the locale module and having active configuration overrides, we run into circular dependencies regarding the language manager service.

Steps to reproduce:

1) enable locale module and a non-English language
2) enable domain_config
3) put a domain-specific override into configuration (I used system.site)
4) Symfony service loader will complain about circular depencency on "language_manager" service

I am not sure what is the best way to avoid this problem but I tried to get around it by postponing the initialization of the domain negotiator and language manager services. Please have a look at it.

Also, entityManager is deprecated, see:
https://www.drupal.org/node/2549139


agentrickard:
Can you cross-reference this with #107 and the problem I was having in https://github.com/agentrickard/domain/tree/107-optional

Does this fix help that move forward?

See also #144 and #145 regarding the entityManager. If this fixes any of those issues, please update and corss-reference with CHANGELOG.md.

domain/src/DomainLoader.php
```diff
@@ -44,7 +44,7 @@ public function loadSchema() {
    * {@inheritdoc}
    */
   public function load($id, $reset = FALSE) {
-    $controller = \Drupal::entityManager()->getStorage('domain');
+    $controller = \Drupal::entityTypeManager()->getStorage('domain');
```
agentrickard, line 48: Shouldn't we be injecting this service?

bjaxelsen, line 48: I don't think this is possible, see #144 (but do try and prove me wrong)

bjaxelsen:
I think this fixes #107.

agentrickard:
I'm on a strategic retreat through Thursday, but plan on reviewing this thoroughly on Friday.

agentrickard:
I'm going to merge this and then make some changes.

Issue #166, opened 2016-01-27
Drupal cache system makes it difficult to serve different homepage for each domain


bjaxelsen:
We need to be able to serve different homepages for each domain. Say we want to show `/node/1` as the homepage of http://localhost1/ and `/node/2` as the homepage of http://localhost2/.

Even though domain config can handle this, we run into problems with route caching in the Drupal cache system:

`\Drupal\Core\Routing\RouteProvider::getRouteCollectionForRequest()` defines cache ID's as `'route:' . $request->getPathInfo() . ':' .  $request->getQueryString()`.

For the home page this boils down to `route:/:`. This cache item will store the internal path to the homepage (say `/node/1`). Depending on the active domain when populating the cache, this can be a domain-specific homepage setting (overwriting system.site configuration).

Unlike the render cache, no cache context is added here. So home page setting for one domain will be active for all other domains as well. So even though we have made a configuration override for each domain, the cache will just serve the first cached domain setting for every domain.

**Possible solutions**

*1) Optional cache context in core for routes*
Extend the cache API:
* Using the kernel.request event any module can add in information to the request about 'route.cache.contexts'. The Request object has a public $attributes property that I think could be used for this.
* Domain config module can add an url context.
* `\Drupal\Core\Routing\RouteProvider::getRouteCollectionForRequest()` can pick up the route cache context and modify the cache ID (similar to the cache context of render caching).

*2) Register proxy path for domain-varied homepage*
* We create a small "domain homepage" module that allows us to set a per-domain homepage. 
* system.site > page.front configuration must be set to a path provided by this module (say `/domain-homepage`) - if not, a notice is shown on `admin/reports/status`
* When enabled, the module adds a homepage setting to `admin/config/domain/edit/{domain}`
* Because we control the caching of this path, it is easy to add in a cache contexts.

What do you think, @agentrickard?

agentrickard:
This is closely related to #123, from a theory standpoint, and I thought we had fixed the issue of caching of site names (which came up earlier).

Does adding the required cache context of `url:site` fix this issue? See https://github.com/agentrickard/domain/blob/8.x-1.x/domain_config/README.md

If it doesn't, I think this belongs in DomainConfig. I would like to keep clean separation of function, so that the core module can be extended more easily.

However, I also worry that the cache context addition in the kernel request might have performance issues, so the smaller module might be wise. I think the real question here is "does this cache problem affect other parts of the config system?"

agentrickard:
[...]

PR #170, opened 2016-01-30
166 homepage

agentrickard:
This is close to working, and based on the approach in #166.

However, URL paths on the generated page are inconsistent. I am running in a subdirectory, and the base_path() is being ignored, breaking the page output.

PR opened so people can look at the code.

See also https://github.com/symfony/symfony/blob/2.8/src/Symfony/Bundle/FrameworkBundle/Controller/Controller.php#L56

### DIFF
--- a/domain_config/src/Tests/DomainConfigHomepageTest.php
+++ b/domain_config/src/Tests/DomainConfigHomepageTest.php
@@ -1,0 +1,70 @@
+<?php
+
+/**
+ * @file
+ * Definition of Drupal\domain\Tests\DomainConfigHomepageTest.
+ */
+
+namespace Drupal\domain_config\Tests;
+
+use Drupal\domain_config\Tests\DomainConfigTestBase;
+
+/**
+ * Tests the domain config system handling of home page routes.
+ *
+ * @group domain_config
+ */
+class DomainConfigHomepageTest extends DomainConfigTestBase {
+
+  public static $modules = array('node', 'views');
+
+  /**
+   * Tests that domain-specific homepage loading works.
+   */
+  function testDomainConfigOverrider() {
+    // Let anon users see content.
+    user_role_grant_permissions(DRUPAL_ANONYMOUS_RID, array('access content'));
+
+    // Configure 'node' as front page.
+    $this->config('system.site')->set('page.front', '/node')->save();
+
+    // No domains should exist.
+    $this->domainTableIsEmpty();
+    // Create four new domains programmatically.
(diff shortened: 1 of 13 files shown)


bjaxelsen:
Works fine on my setup, thanks!

However, I can't make the `DomainConfigHomepageTest::testDomainConfigOverrider()` test pass.

The "Proper home page loaded." test fails. This is due to this error:

```
Undefined index: four_localhost
Drupal\domain_config\Tests\DomainConfigHomepageTest->testDomainConfigOverrider()
```

[test.html.zip](https://github.com/agentrickard/domain/files/112673/test.html.zip)

domain_config/src/Tests/DomainConfigHomepageTest.php
```diff
@@ -0,0 +1,70 @@
...
+      $expected = $domain->getPath() . $homepages[$domain->id()];
+      $expected_home = $this->drupalGet($expected);
+      $this->assertTrue($home == $expected_home, 'Proper home page loaded.');
+    }
+  }
+
+  /**
+   * Returns the expected homepage paths for each domain.
+   */
+  private function getHomepages() {
```
agentrickard: Rewrite this part to use the keys from actual domains.

agentrickard:
The logic in getHomepages() hardcodes machine names. That needs to be changed.

agentrickard:
I suspect the failures that I get are because I am running in a subdirectory. That worries me, because it means that the code isn't working in the same way as core.

Did you see any issues with active link highlighting or breadcrumbs?

bjaxelsen:
In the test result? No, I don't think so (like a written breadcrumb?), have a look here:
[test_results.docx](https://github.com/agentrickard/domain/files/112749/test_results.docx)

I haven't tried running in a subpath.

agentrickard:
I had problems with breadcrumbs and links when using the UI, not in tests.

For tests, I suspect that none of the Domain Config tests will pass for you, since the schema files (the config overrides) are keyed to example_com not localhost. We could fix that by changing this part of the test domain creation routine:

```
      // Create a new domain programmatically.
      $values = array(
        'hostname' => $hostname,
        'name' => $name,
        'id' => \Drupal::service('domain.creator')->createMachineName($hostname),
      );
```

We should use example_com (or localhost) consistently. domainCreateTestDomains() handles different hosts properly (otherwise none of the tests would pass), but it is mismatching on the machine name of the domain.

agentrickard:
Pushed a commit that should help with that specific test fail.

PR #172, opened 2016-02-02
116 help text to domain overview screen

artreaktor:
Drupal caches help text, so if the default domain was changed it provides obsolete domain name. 
Probably we need to rebuild cache on DomainListBuilder submitForm. Need suggestion to overcome this.

### DIFF
--- a/domain/src/DomainForm.php
+++ b/domain/src/DomainForm.php
@@ -117,7 +117,13 @@ class DomainForm extends EntityForm {
     else {
       drupal_set_message($this->t('Domain record updated.'));
     }
+
+    if ($domain->isDefault()) {
+      \Drupal\Core\Cache\Cache::invalidateTags(array('rendered'));
+    }
+
     $domain->save();
+
     $form_state->setRedirect('domain.admin');
   }
 
--- a/domain/src/DomainForm.php
+++ b/domain/src/DomainForm.php
@@ -117,13 +117,8 @@ class DomainForm extends EntityForm {
     else {
       drupal_set_message($this->t('Domain record updated.'));
     }
-
-    if ($domain->isDefault()) {
-      \Drupal\Core\Cache\Cache::invalidateTags(array('rendered'));
-    }
-
+    \Drupal\Core\Cache\Cache::invalidateTags(array('rendered'));
     $domain->save();
-
     $form_state->setRedirect('domain.admin');
   }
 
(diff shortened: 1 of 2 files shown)


agentrickard:
There should be a cache tag responsible for the help text, so clearing that should work.

agentrickard:
It is acceptable if we don't name the default domain in the help text, but instead point to the list below. 

What needs some checking is whether changing the default domain throws off the cache behavior in other instances.

artreaktor:
I noticed that help text is cached with the following tags:
block_view config:block.block.seven_help rendered
so, there's no domain-specific tag there

Probably we need to implement
\Drupal\Core\Entity\EntityInterface::getCacheTagsToInvalidate()
in \Drupal\domain\DomainInterface
in order to have ourown domain-related tag

Also we can? (not sure) invalidate it by [url] in cid:
entity_view:block:seven_help:
[languages:language_interface]=en:
[theme]=seven:
**[url]=http://regofix.dd/admin/config/domain:**[user.permissions]=e30e625ddec9ded6bc63dddee5042e0b1a44da0e24fa1e2ef0808730d987370

What do you think?

agentrickard:
We should be able to invalidate with url:site, which is what we use rather than providing our own cache context.

See https://github.com/agentrickard/domain/blob/8.x-1.x/domain_config/README.md

artreaktor:
didn't find nothing more specific than just
\Drupal\Core\Cache\Cache::invalidateTags(array('rendered'));

agentrickard:
Try that. If it fails, we can remove the dynamic text from this help string. 

There is drupal_clear_all_caches() as a last resort.

artreaktor:
\Drupal\Core\Cache\Cache::invalidateTags(array('rendered'));
works

domain/src/DomainForm.php
```diff
@@ -117,7 +117,13 @@ public function save(array $form, FormStateInterface $form_state) {
     else {
       drupal_set_message($this->t('Domain record updated.'));
     }
+
+    if ($domain->isDefault()) {
```
agentrickard: I suspect we want to clear the render cache when any domain changes, just to be safe.

We may need to do that for aliases records as well.

agentrickard:
The cache invalidation should happen at the entity level, not in the save routine.  It turns out that the base entity class does this already:

```
  /**
   * {@inheritdoc}
   */
  public function postSave(EntityStorageInterface $storage, $update = TRUE) {
    $this->invalidateTagsOnSave($update);
  }
```

See `core/lib/Drupal/Core/Entity/Entity.php`
We can implement this method in our Domain and DomainAlias classes, and check that we are invalidating the proper tag set.

artreaktor:
Created new pull request #184

Issue #160, opened 2016-01-22
Deprecated methods in DomainAliasController


agentrickard:
Remove the old methods.

artreaktor:
Unable to save domain alias: DB crashes on each try, and site becomes unavailable (exceed timeout limit). Only re-import database dump helps.
I compared data passed to save() for Domain and DomainAlias and found smth interesting:
**$domain->domain_id** is integer, but
**$alias->domain_id** is string (parent domain mashine name)
Need help to deal with this

agentrickard:
It's a dumb error in the code. We should be using the id() of the domain (a string). $domain-domain_id is only used for node access. 

See domainAlaisCreateTeatAlias().

agentrickard:
No, that's not it. I think it's somwthing in the redirect handler. 

Unfortunately, my internet is out, so I can't share what I'm seeing.

artreaktor:
According to mysql query log query
SELECT cid, data, created, expire, serialized, tags, checksum FROM cache_config WHERE cid IN ( 'domain_alias.alias.china_regofix_tk' ) ORDER BY cid
repeated endlessly

agentrickard:
There is an infinite loop being invoked by domain_requwst_alter().

artreaktor:
not sure I understand how to deal with this

artreaktor:
I experimented with
DomainAliasController::addAlias()
with smth like
$values['domain_id'] = 5; 
(passing int, not string)
This solves the looping issue, but obviosly brings another ones. 
However, the main issue can be caused by wrong domain_id type, I guess

agentrickard:
I don't think it has anything to do with the IDs. Those all seem to be fine (and are handled behind the scenes).

In that code sample, though, 'domain_id' should be a string (e.g. example_com) that is the machine name of the referencing domain. ($domain->id()).

In my tests, it was a runaway load call. I'll push the *hotfix-alias* branch in a minute for you to test.

agentrickard:
See #173 or the *hotfix-alias* branch for testing.

agentrickard:
Fixed.

PR #173, opened 2016-02-04
Fixes infinite loop in domain loading.

agentrickard:
See #160


artreaktor:
Seems to work now, thanks! Added #177 alias filter to display only relevant aliases

PR #176, opened 2016-02-04
Invalidate render cache on Alias save

artreaktor:


domain_alias/src/DomainAliasForm.php
```diff
@@ -89,6 +89,7 @@ public function save(array $form, FormStateInterface $form_state) {
     else {
       drupal_set_message($this->t('Domain alias updated.'));
     }
+    \Drupal\Core\Cache\Cache::invalidateTags(array('rendered'));
```
agentrickard, line 92: I like to have a `use` declaration at the top if we're going to call a static. But I thing this should be called from the service handler, not by directly calling the class.

e.g.

`\Drupal::services('cache')`

agentrickard:
See https://www.palantir.net/blog/d8ftw-customizing-your-back-end and the list of services in `/core/core.services.yml`

We probably want the `cache_tags.invalidator` service.

agentrickard:
See https://github.com/agentrickard/domain/pull/172#issuecomment-180384361

PR #177, opened 2016-02-04
Hotfix alias: filter aliases by selected domain

### LINKED ISSUE #178: Aliases are not filtered by selected domain_id
admin/config/domain/alias/example_com shows full aliases list.

artreaktor:
Fixes #178  The alias list displays irrelevant items despite the selected domain

### DIFF
--- a/domain_alias/src/DomainAliasListBuilder.php
+++ b/domain_alias/src/DomainAliasListBuilder.php
@@ -9,7 +9,6 @@ namespace Drupal\domain_alias;
 
 use Drupal\Core\Config\Entity\ConfigEntityListBuilder;
 use Drupal\Core\Entity\EntityInterface;
-use Symfony\Cmf\Component\Routing;
 
 /**
  * User interface for the domain alias overview screen.
--- a/domain_alias/src/DomainAliasListBuilder.php
+++ b/domain_alias/src/DomainAliasListBuilder.php
@@ -47,11 +47,7 @@ class DomainAliasListBuilder extends ConfigEntityListBuilder {
    * {@inheritdoc}
    */
   public function render() {
-
-    $current_path = \Drupal::service('path.current')->getPath();
-    $path_args = explode('/', $current_path);
-    $domain_id = array_pop($path_args);
-
+    $domain_id = \Drupal::routeMatch()->getParameter('domain');
     $build = array(
       '#theme' => 'table',
       '#header' => $this->buildHeader(),
--- a/domain_alias/src/DomainAliasListBuilder.php
+++ b/domain_alias/src/DomainAliasListBuilder.php
@@ -9,6 +9,7 @@ namespace Drupal\domain_alias;
 
 use Drupal\Core\Config\Entity\ConfigEntityListBuilder;
 use Drupal\Core\Entity\EntityInterface;
+use Symfony\Cmf\Component\Routing;
 
 /**
  * User interface for the domain alias overview screen.
@@ -47,6 +48,11 @@ class DomainAliasListBuilder extends ConfigEntityListBuilder {
    * {@inheritdoc}
    */
   public function render() {
+
+    $current_path = \Drupal::service('path.current')->getPath();
+    $path_args = explode('/', $current_path);
+    $domain_id = array_pop($path_args);
+
     $build = array(
       '#theme' => 'table',
       '#header' => $this->buildHeader(),
@@ -54,8 +60,10 @@ class DomainAliasListBuilder extends ConfigEntityListBuilder {
       '#empty' => $this->t('No aliases have been created for this domain.'),
     );
     foreach ($this->load() as $entity) {
-      if ($row = $this->buildRow($entity)) {
-        $build['#rows'][$entity->id()] = $row;
+      if ($entity->getDomainId() == $domain_id) {
+        if ($row = $this->buildRow($entity)) {
+          $build['#rows'][$entity->id()] = $row;
+        }
       }
     }
     return $build;


domain_alias/src/DomainAliasListBuilder.php
```diff
@@ -47,6 +47,11 @@ public function buildRow(EntityInterface $entity) {
    * {@inheritdoc}
    */
   public function render() {
+
+    $current_path = \Drupal::service('path.current')->getPath();
```
agentrickard: We really should be able to inject this variable into the route.

agentrickard: And this filtering should happen at the getEntityIds() level, which can be overridden by this class.

I can't see immediately how to upcast the $domain from the route without rewriting this as a Controller. 

But you should not parse the url, use the notes here -- https://www.drupal.org/node/2310429

agentrickard:
This needs a proper issue.

agentrickard:
Here's a long conversation on IRC that tries to dig into the solution.

```
agentrickard    Is it possible to pass a Route parameter (e.g. {node} to an EntityListBuilder? 
agentrickard    it doesn't seem to allow for that
larowlan    agentrickard: hey, so do you have _entity_list in your routing.yml def?
agentrickard    yes
larowlan    agentrickard: That is 'enhanced' by EntityRouteEnhancer, see \Drupal\Core\Entity\Enhancer\EntityRouteEnhancer::enhanceEntityList as you can see that uses the EntityListController::listing for controller
larowlan    agentrickard: you would need to add a new enhancer to swap out EntityListController::listnig with something of your own, that had a {node}/$node argument and then handle building the list builder/render() call differently
agentrickard    that seems onerous
larowlan    agentrickard: you could override the render() method in your list controller as long as you did it in a way that was consistent with the interface (ie $node = NULL as argument, not just $node)
agentrickard    right
agentrickard    I was thinking it should override getEntityIds() actually, since that's where the query is
larowlan    agentrickard: if you don't want to add a new enhancer, don't use _entity_list, instead hard-code the controller in your defaults
agentrickard    and I can pull the node param off the route, but would prefer to pass it 
agentrickard    that means I have to write a new controller that mirrors the features of the EntityList
agentrickard    which also seems onerous
larowlan    agentrickard: all _entity_list does is 'enhance' the routing.yml bits as per \Drupal\Core\Entity\Enhancer\EntityRouteEnhancer::enhanceEntityList, so if you put those bits in your routing,yml, you don't need _entity_list
larowlan    agentrickard: the controller is EntityListController::listing, its a one-liner
agentrickard    i am just not following
agentrickard    is there any code that uses that pattern?
larowlan    agentrickard: if you remove _entity_list from your routing.yml and instead use your own _controller that copies what EntityListController::listing does, but instead passes a $node to render() method, and then your list controller has function render($node = NULL), saving $node as a class property for access in getEntityIds, that should work
agentrickard    and isn't EntityListContoller now EntityListBuilder? 
larowlan    agentrickard: no EntityListController is the routing glue between routing system and EntityListBuilder
larowlan    agentrickard: if you use _entity_list in your routing.yml, it uses EntityListController::listing to call the EntityListBuilder
larowlan    agentrickard: \Drupal\Core\Entity\Enhancer\EntityRouteEnhancer::enhanceEntityList is where the routing.yml entry takes the _entity_list and builds the correct controller
agentrickard    fascinating
agentrickard    and what do I do with my entity annotation? pass the new controller?
larowlan    agentrickard: see \Drupal\Core\Entity\Enhancer\EntityRouteEnhancer::enhanceEntityList
larowlan    agentrickard: in that case $defaults is the _defaults hunk in routing.yml
larowlan    agentrickard: so it puts the _entity_list value in _defaults entity_type
larowlan    agentrickard: and nominates the controller
agentrickard    I see, so what about the list_builder in the entity annotation? Replace with the new controller? Remove?
larowlan    agentrickard: no keep that
larowlan    agentrickard: that is used by EntityListController::listing (which you are going to sub-class or replace to support a $node param)
larowlan    agentrickard: the ->getListBuilder() call in EntityListController::listing  needs that annotation
```

agentrickard:
Here's a patch that shows how that might work. I'm in a rush right now.

```
diff --git a/domain_alias/domain_alias.routing.yml b/domain_alias/domain_alias.routing.yml
index 4bb16d2..35a68a6 100644
--- a/domain_alias/domain_alias.routing.yml
+++ b/domain_alias/domain_alias.routing.yml
@@ -1,7 +1,7 @@
 domain_alias.admin:
   path: '/admin/config/domain/alias/{domain}'
   defaults:
-    _entity_list: 'domain_alias'
+    _controller: '\Drupal\domain_alias\Controller\DomainAliasController::listing'
     _title: 'Domain aliases'
     _title_context: 'With modules'
   requirements:
diff --git a/domain_alias/src/Controller/DomainAliasController.php b/domain_alias/src/Controller/DomainAliasController.php
index 6dfc359..d0040a4 100644
--- a/domain_alias/src/Controller/DomainAliasController.php
+++ b/domain_alias/src/Controller/DomainAliasController.php
@@ -9,6 +9,7 @@ namespace Drupal\domain_alias\Controller;

 use Drupal\domain\DomainInterface;
 use Drupal\domain\Controller\DomainControllerBase;
+use Drupal\domain_alias\DomainAliasListBuilder;

 /**
  * Returns responses for Domain Alias module routes.
@@ -34,4 +35,20 @@ class DomainAliasController extends DomainControllerBase {
     return $this->entityFormBuilder()->getForm($alias);
   }

+  /**
+   * Provides the listing page for aliases.
+   *
+   *   An domain record entity.
+   *
+   * @return array
+   *   A render array as expected by drupal_render().
+   */
+  public function listing(DomainInterface $domain) {
+    $type = \Drupal::entityTypeManager()->getType('domain_alias');
+    $storage = \Drupal::entityTypeManager()->getStorage('domain_alias');
+    $list = new DomainAliasListBuilder($type, $storage);
+    return $list->render();
+  }
+
 }
diff --git a/domain_alias/src/DomainAliasListBuilder.php b/domain_alias/src/DomainAliasListBuilder.php
index 621b60a..c5b8d64 100644
--- a/domain_alias/src/DomainAliasListBuilder.php
+++ b/domain_alias/src/DomainAliasListBuilder.php
@@ -2,7 +2,7 @@

 /**
  * @file
- * Contains \Drupal\domain_alias\Form\DomainAliasListBuilder.
+ * Contains \Drupal\domain_alias\DomainAliasListBuilder.
  */
```

agentrickard:
I did it properly. See #179

PR #179, opened 2016-02-04
Rewire the domain alias controller.

agentrickard:
Change the behavior of the DomainAliasListBuilder to be domain sensitive.

### DIFF
--- a/domain_alias/src/Controller/DomainAliasController.php
+++ b/domain_alias/src/Controller/DomainAliasController.php
@@ -9,7 +9,6 @@ namespace Drupal\domain_alias\Controller;
 
 use Drupal\domain\DomainInterface;
 use Drupal\domain\Controller\DomainControllerBase;
-use Drupal\domain_alias\DomainAliasListBuilder;
 
 /**
  * Returns responses for Domain Alias module routes.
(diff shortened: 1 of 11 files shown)


artreaktor:
The object passed in DomainAliasController::listing
is of type DomainAliasListBuilder, not Domain
Getting Fatal error: Call to a member function id() on null in \DomainAliasListBuilder.php on line 77

agentrickard:
Do you know how to fix that? 

See https://secure.php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration

agentrickard:
I also don't get that error. Did you rebuild the container? `drush cr`.

The route was rebuilt, and the variable should be passed from the routing call.

```
domain_alias.admin:
  path: '/admin/config/domain/alias/{domain}'
  defaults:
    _controller: '\Drupal\domain_alias\Controller\DomainAliasController::listing'
    _title: 'Domain aliases'
    _title_context: 'With modules'
  requirements:
    _permission: 'administer domains'
```

{domain} here should automatically be passed to listing().

domain_alias/src/Controller/DomainAliasController.php
```diff
@@ -34,4 +35,21 @@ public function addAlias(DomainInterface $domain) {
...
+   * @param \Drupal\domain\DomainInterface $domain
+   *   An domain record entity.
+   *
+   * @return array
+   *   A render array as expected by drupal_render().
+   */
+  public function listing(DomainInterface $domain) {
+    $type = \Drupal::entityTypeManager()->getDefinition('domain_alias');
+    $storage = \Drupal::entityTypeManager()->getStorage('domain_alias');
+    $list = new DomainAliasListBuilder($type, $storage);
```
agentrickard: We may be able to just call the service directly. e.g.

```
$list = $this->entityManager()->getListBuilder($entity_type);
```

See EntityListController::listing().

PR #183, opened 2016-02-09
Add render cache invalidation on Domain save

### LINKED ISSUE #182: Invalidate cache on Domain save

artreaktor:
Fix #182

### DIFF
--- a/domain/src/Entity/Domain.php
+++ b/domain/src/Entity/Domain.php
@@ -310,6 +310,14 @@ class Domain extends ConfigEntityBase implements DomainInterface {
   /**
    * {@inheritdoc}
    */
+  public function postSave(EntityStorageInterface $storage, $update = TRUE) {
+    parent::postSave($storage, $update);
+    \Drupal::service('cache_tags.invalidator')->invalidateTags(array('rendered'));
+  }
+
+  /**
+   * {@inheritdoc}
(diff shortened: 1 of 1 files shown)


agentrickard:
So this one really made me track through the call stack, and it turns out that the list of cache tags is defined as part of the entity annotation.

See:
- /Drupal/Core/Entity/Entity::InvalidateTagsOnSave()
- /Drupal/Core/Entity/EntityType::getListCacheTags()
- /Drupal/Core/Entity/EntityType::__construct()

And then, as an example, the annotation in \Drupal\node\Entity\Node.

So we just need to add the tags to our entity annotation:
-   list_cache_contexts = { "renderer" },

However, we can pass an array of contexts here, and I wonder which ones are required.  The best way to test this is probably to remove the drupal_clear_all_caches() statement from the DomainInactiveTest.

I think we should clear 'url:site', 'renderer', and possibly 'user:node_grants'.

agentrickard:
See https://www.drupal.org/developing/api/8/cache/contexts for a list of contexts.

url:site maps to domain, which is why we don't provide our own cache context.

agentrickard:
Looks like the entity annotation isn't relevant here.

agentrickard:
Revised as #246

PR #184, opened 2016-02-09
Add hepl text to domain overview screen

### LINKED ISSUE #116: Add help text to domain overview screen
The overview form (admin/config/domain) has no text explaining the page.  We can pull copy from the Drupal 7 `domain_overview_form` function, but it could use a rewrite for the new options.

The D8 form code is in `\Drupal\domain\DomainListBuilder`

artreaktor:
Fix #116


agentrickard:
And let's deal with the caching issues in follow-up? Or is that coming here, too?

agentrickard:
Thanks!

PR #191, opened 2016-02-15
Add current domain condition

### LINKED ISSUE #190: Add current domain condition plugin

artreaktor:
Fix #190

### DIFF
--- a/domain/src/Plugin/Condition/CurrentDomain.php
+++ b/domain/src/Plugin/Condition/CurrentDomain.php
@@ -1,0 +1,97 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\domain\Plugin\Condition\CurrentDomain.
+ */
+
+namespace Drupal\domain\Plugin\Condition;
+
+use Drupal\Core\Condition\ConditionPluginBase;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Plugin\Context\ContextDefinition;
+
+/**
+ * Provides a 'Current domain' condition.
+ *
+ * @Condition(
+ *   id = "current_domain",
+ *   label = @Translation("Current domain"),
+ *   context = {
+ *     "domain" = @ContextDefinition("entity:domain", label = @Translation("Domain"))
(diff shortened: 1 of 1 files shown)


domain/src/Plugin/Condition/CurrentDomain.php
```diff
@@ -0,0 +1,97 @@
...
+use Drupal\Core\Plugin\Context\ContextDefinition;
+
+/**
+ * Provides a 'Current domain' condition.
+ *
+ * @Condition(
+ *   id = "current_domain",
+ *   label = @Translation("Current domain"),
+ *   context = {
+ *     "domain" = @ContextDefinition("entity:domain", label = @Translation("Domain"))
```
agentrickard: Malformed annotation

domain/src/Plugin/Condition/CurrentDomain.php
```diff
@@ -0,0 +1,97 @@
...
+   */
+  public function submitConfigurationForm(array &$form, FormStateInterface $form_state) {
+    $this->configuration['current_domain'] = array_filter($form_state->getValue('current_domain'));
+    parent::submitConfigurationForm($form, $form_state);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function summary() {
```
agentrickard, line 67: How does this interact with domain.js?

agentrickard:
I don't understand the use-case for this code. Where is it visible in the UI? Why is the existing ContextProvider not sufficient?

artreaktor:
What's wrong with annotation? I used Drupal\domain\Plugin\Condition\Domain.php as an example.

agentrickard:
I see. That annotation is actually a method signature, so one line is ok.

See Drupal\node\Plugin\Condition\NodeType

agentrickard:
OK, so we don't need a next Condition handler, we need to make sure that the existing one \Drupal\domain\Plugin\Condition\Domain functions properly for Blocks and Panels, yes?

artreaktor:
+1

PR #194, opened 2016-02-16
Improve domain context global visibility

### LINKED ISSUE #193: Improve domain context global visibility
I compared globally available entities implementation and it seems that we have to implement ContainerFactoryPluginInterface.
Look at \Drupal\system\Plugin\Condition\CurrentThemeCondition, for instance.

artreaktor:
Fix #193

### DIFF
--- a/domain/src/Plugin/Condition/Domain.php
+++ b/domain/src/Plugin/Condition/Domain.php
@@ -9,7 +9,9 @@ namespace Drupal\domain\Plugin\Condition;
 
 use Drupal\Core\Condition\ConditionPluginBase;
 use Drupal\Core\Form\FormStateInterface;
-use Drupal\Core\Plugin\Context\ContextDefinition;
+use Drupal\domain\DomainNegotiator;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
  * Provides a 'Domain' condition.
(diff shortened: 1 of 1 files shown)


domain/src/Plugin/Condition/Domain.php
```diff
@@ -89,9 +124,19 @@ public function evaluate() {
     if (empty($domains) && !$this->isNegated()) {
       return TRUE;
     }
-    $domain = $this->getContextValue('domain');
+    $domain = $this->domainNegotiator->getActiveDomain();
```
agentrickard, line 127: This feels wrong to me. Doesn't the Content handler pass this value?

domain/src/Plugin/Condition/Domain.php
```diff
@@ -89,9 +124,19 @@ public function evaluate() {
...
     // NOTE: The block system handles negation for us.
     return (bool) in_array($domain->id(), $domains);
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheContexts() {
+    $contexts = parent::getCacheContexts();
+    $contexts[] = 'domain';
```
agentrickard, line 138: We use the `url.site` cache context, not a domain context, since url.site covers our needs.

agentrickard:
This change breaks the Block visibility handler.

agentrickard:
I take that back. trying to run $this->getContextValue('domain') breaks everything.

domain/src/Plugin/Condition/Domain.php
agentrickard, line 127: It tests out ok. So let's leave this in. It's how the Theme Condition works.

agentrickard:
Well, this change causes existing tests to fail.

```
Drupal\Component\Plugin\Exception\ContextException: The domain context is not a valid context. in Drupal\Component\Plugin\ContextAwarePluginBase->getContextDefinition() (line 84 of /Applications/MAMP/htdocs/drupal-8/core/lib/Drupal/Component/Plugin/ContextAwarePluginBase.php). 
```

artreaktor:
Can't figure out how to solve this. Any hints?

agentrickard:
Not a one. The problem with the test is that we removed the Context from the plugin annotation. If we add the Context to the annotation, the code breaks.

I'm trying to schedule some time with EclipseGc (who wrote a lot of this) to try to figure out what is going on.

I know why the test breaks, and maybe the test needs to be rewritten. What I don't understand is why we can't have that annotation there. The active domain should be a default context, like the current user.

artreaktor:
I inserted

```
    dpm($plugin_id);
    dpm($configuration);
```

at the beginning of the
`ContextAwarePluginBase::__construct()`

Domain context appears once among others after cache clearing and disappears on page refresh, so probably we have caching issue

Also I'm curious why the CurrentDomainContext provider prefixed with block._, not domain._ in the domain.services.yml. However, I didn't noticed any dofference changing it to domain.*

artreaktor:
Also $configuration of the domain doesn't have `provider` key defined

artreaktor:
Having 
`$domain = $this->domainNegotiator->getActiveDomain();`
I'm getting 
**Fatal error: Call to a member function id() on a non-object in /modules/domain/domain/src/Plugin/Condition/Domain.php on line 131**
and
**Container cannot be saved to cache.**
in watchdog log
when trying to run /update.php

agentrickard:
It was prefixed block.\* because that's what all the core examples did.

$connfiguration key in which part of the code?

artreaktor:
`dpm($configuration);`
at the beginning of the
`ContextAwarePluginBase::__construct()`

agentrickard:
Right, I meant "where in the code do we set that property?"

BTW, if you use Devel, enable the new kint module and use kint() instead of dpm(). It's nicer and shows you all the object details.

artreaktor:
If you mean `provider` property, I don't know where it should be set, just noticed it present in all context configurations, except domain
Thank you for the suggestion, I have kint enabled and it looks like dpm() output already wrrapped by kint, not old-good krumo :)

agentrickard:
Testing in #197.

PR #195, opened 2016-02-17
Add .travis.yml

zerolab:
This PR adds a .travis.yml configuration to allow running simpletest tests on travis-ci.org
Example output: https://travis-ci.org/zerolab/domain/jobs/109740201

I am aware of `drush generate-domains` and `drush domain-test` and I have tried an alternative approach to the .travis.yml file ([diff](https://github.com/agentrickard/domain/compare/8.x-1.x...zerolab:travis-drush-tests)), however it requires sudo and that is only supported on the legacy Travis platform which is **much** slower.

I found having Travis on a GitHub project much better in terms of workflow as it provides a glimpse in the PR changes without having to check them locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,85 @@
+language: php
+sudo: false
+
+php: [5.5, 5.6, 7.0, hhvm]
+
+matrix:
+  fast_finish: true
+  allow_failures:
+    - php: hhvm
+
+addons:
+  hosts:
+    - one.localhost
+    - two.localhost
+    - three.localhost
+    - four.localhost
+    - five.localhost
+
+cache:
+  directories:
+    - $HOME/.composer/cache/files
+
+env:
+  - DRUPAL=8.0.x
+  - DRUPAL=8.1.x
+
+mysql:
+  database: domain
+  username: root
+  encoding: utf8
+
+notifications:
+  email: false
+
+before_install:
+  # Add composer's global bin directory to the path
+  # see: https://github.com/drush-ops/drush#install---composer
+  - export PATH="$HOME/.composer/vendor/bin:$PATH"
+
+  # Remove Xdebug. Not an issue for PHP 7.
+  - phpenv config-rm xdebug.ini || true
+
+  - composer self-update
+
+install:
+  # Install Drush.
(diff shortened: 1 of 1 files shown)


agentrickard:
This is awesome. Two questions / caveats:
- I can't really help with this right now.
- How does Travis handle the subdomains required for testing. e.g.
  `GET http://two.localhost/ returned 0 (0 bytes).`

I see the answer to 2 in the config file, but didn't see it help in the test output.

zerolab:
Thank you for the feedback.

Not a problem re: help. I will dig into it again tonight. Should not be long now.
Will update with more details on Travis setup once this is ready

agentrickard:
Question: Is there a benefit or preference for Travis over Circle? 

At Palantir, we recently switched to Circle.

zerolab:
Other than Travis being more widespread and widely used, I don't know.
At Torchbox we started using a self-hosted [Drone](https://github.com/drone/drone) setup, alongside Travis.

I do not see why not run both Travis and Circle and decide over time which one is faster and provides more benefits.

agentrickard:
Given that you're taking this on, Travis is fine.

agentrickard:
I went ahead and authorized this project on Travis.

https://travis-ci.org/profile/agentrickard

What next?

agentrickard:
If I read that correctly, most of the test fails are a cascade from a failed admin login.

zerolab:
Was looking at that. The issue is the PR missing running the built-in PHP server.
Testing a few variations and will push it back to this branch shortly.

zerolab:
I think we're ready.

You could add a badge to REAME.md for main branch, if you wish to:

```
[![Build Status](https://travis-ci.org/agentrickard/domain.svg?branch=8.x-1.x)](https://travis-ci.org/agentrickard/domain)
```

agentrickard:
You Rock!

zerolab:
Thank you.
Took some trial and error, but we're there. I think it is ready to merge so 8.x-1.x has it. Any PR in progress could be rebased on 8.x-1.x so it gets the travis integration.

Will look at making tests run on hhvm tonight.

PR #197, opened 2016-02-20
194 context

agentrickard:
Extends the work by  artreaktor on #194.


agentrickard:
OK, this seems to be working as desired.  What I had to do is update the Context annotation to make required = FALSE. (See `\Drupal\Core\Annotation\ContextDefinition`). 

That allows us to override Page Manager's desire to skip the plugin. 

I think the plugin keys are named correctly, when I didn't use `entity:domain`, the value was being overwritten by CTools, I think.

Tests are passing locally.

PR #209, opened 2016-03-30
Follow up on serving different home page

bjaxelsen:
I've made an alter of Drupal core's route handling thrugh implementation of \Drupal\Core\DependencyInjection\ServiceProviderBase::alter() and made changes to the test of delivering different home pages.

### DIFF
--- a/domain_config/src/DomainConfigServiceProvider.php
+++ b/domain_config/src/DomainConfigServiceProvider.php
@@ -1,0 +1,25 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\domain_config\DomainConfigServiceProvider.
+ */
+
+namespace Drupal\domain_config;
+
+use Drupal\Core\DependencyInjection\ContainerBuilder;
+use Drupal\Core\DependencyInjection\ServiceProviderBase;
+
+/**
+ * Overrides the router.route_provider service to point to our customized one.
+ */
+class DomainConfigServiceProvider extends ServiceProviderBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function alter(ContainerBuilder $container) {
+    $definition = $container->getDefinition('router.route_provider');
+    $definition->setClass('Drupal\domain_config\Routing\DomainRouteProvider');
+  }
+}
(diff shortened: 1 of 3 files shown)


bjaxelsen:
See https://www.drupal.org/node/2662196 and #166

domain_config/src/DomainConfigServiceProvider.php
```diff
@@ -0,0 +1,25 @@
...
+ * Overrides the router.route_provider service to point to our customized one.
+ */
+class DomainConfigServiceProvider extends ServiceProviderBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function alter(ContainerBuilder $container) {
+    $definition = $container->getDefinition('router.route_provider');
+    $definition->setClass('Drupal\domain_config\Routing\DomainRouteProvider');
```
Crell: In PHP 5.5, it's safe and I would argue preferred to use the class-constant-form for a class name.  That is:

``` php
setClass(DomainRouteProvider::class);
```

And then also add a "use" statement above for that class. That way your IDE can recognize it as a class name, spell check it, find it in refactor commands, etc. etc.

bjaxelsen: thanks @Crell - I did not know this before now :-)

domain_config/src/Tests/DomainConfigHomepageTest.php
```diff
@@ -48,10 +50,22 @@ function testDomainConfigHomepage() {
     ));
     $homepages = $this->getHomepages();
     foreach ($domains as $domain) {
+      // Clone the site configuration to set individual home pages if different
+      // than default
+      if ($homepages[$domain->id()] != 'node') {
```
agentrickard: These should be picked up from `/tests/modules/domain_config_test/config/install`. Is that not happening?

agentrickard:
That Travis failure looks like network interruption. Re-running tests.

I like this approach, as it bypasses the need for a core patch.

bjaxelsen:
@agentrickard I had not seen the yml files in the test module, so no need to set the domain homepages again in the DomainConfigHomepageTest - I've cleaned this up.

agentrickard:
Great. Thank you.

PR #213, opened 2016-04-21
Fix SQL exception raised by domain_access_node_grants

claire-remy:
domain_access_node_grants raises "Drupal\Core\Database\DatabaseExceptionWrapper: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL". 
This PR fixes the grants syntax error.


agentrickard:
How is this not caught by our test coverage?

PR #215, opened 2016-04-22
[WIP] Catch all test fails in Travis CI and fix failing tests

zerolab:
Updates .travis.yml with additional `after_script` checking.

Should catch all failing tests.
The reason #213 was not caught is that I missed the submodule from tests when the initial Travis PR went in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,4 +76,4 @@ before_script:
   - until curl -s localhost:8080; do true; done > /dev/null
 
 script:
-  - php core/scripts/run-tests.sh --verbose --color --concurrency 4 --php `which php` --url http://localhost:8080 "domain,domain_access,domain_alias,domain_config,domain_source" | tee /tmp/domain-test-results.txt
+  - php core/scripts/run-tests.sh --verbose --color --concurrency 4 --php `which php` --url http://localhost:8080 "domain,domain_access,domain_alias,domain_config,domain_source"
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ before_install:
 
 install:
   # Install Drush.
-  - composer global require drush/drush:dev-master
+  - composer global require drush/drush:8.*
   - phpenv rehash
 
   # Create database.
--- a/domain_alias/src/Tests/DomainAliasNegotiatorTest.php
+++ b/domain_alias/src/Tests/DomainAliasNegotiatorTest.php
@@ -7,8 +7,6 @@
 
 namespace Drupal\domain_alias\Tests;
 
-use Drupal\domain\DomainInterface;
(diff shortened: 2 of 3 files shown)


agentrickard:
Not sure why this one is failing, when the last two I did this morning worked.

https://travis-ci.org/agentrickard/domain/builds

zerolab:
Aah, got to the bottom of it!

Take https://travis-ci.org/agentrickard/domain/jobs/125022504 for example.

```
The command "php core/scripts/run-tests.sh --verbose --color --concurrency 4 --php `which php` --url http://localhost:8080 "domain,domain_access,domain_alias,domain_config,domain_source" | tee /tmp/domain-test-results.txt" exited with 0.
```

Running the command without piping the tests output to `tee` returns the correct status (i.e. 0 for success, 1 for fail)

My initial PR was piping the results to `/tmp/domain-test-results.tx` then grepping it for errors/exceptions/fails. That was due to the fact that in Drupal < 8, `run-script.php` was reporting the wrong exit status.

As it took a few tries to get all the settings right, I squashed the unneeded commits and suspect the `| tee` part sneaked in at that point.

This should now match what I am seeing when running the tests locally.

agentrickard:
All tests just ran clean for me locally (on 8.1.x).

#213 is an interesting case, because I run tests locally periodically, and they never caught it. That suggests insufficient test coverage.

zerolab:
I think this PR needs more work. https://travis-ci.org/agentrickard/domain/jobs/125036957 failed.
This matches what I see on 8.0.x, will try later today on 8.1.x

agentrickard:
Could be that your fork is out of date?  I did commit a PR today.

https://github.com/agentrickard/domain/commits/8.x-1.x

agentrickard:
Was this corrected by #219?

zerolab:
Not entirely, no.

#219 was necessary due to a recent change in drush:dev which prevented the tests from running at all.

This (wip) PR aims to fix an oversight in the initial Travis setup and fix failing tests. The oversight is likely to have happened in a rebase/commit squash as it took plenty of tries to get the right combination to make Travis happy. The change that slipped through was piping the test results to `tee` which saved them in a file. Initially `.travis.yml` had an `after_script` clause that was parsing the file. Then I discovered that unlike previous versions the Drupal 8 `run-tests.sh` actually returns proper exit codes, so there was no need for the `after_script`.

With that aside, if you check commit e2d6571, `domainAliasCreateTestAlias()` was creating the wrong domain id. And because tests run on port 8080, the resulting id was something along the lines of `__localhost:8080` instead of `__localhost_8080` resulting in an exception in `ConfigBase::validate()`.

I need to do more work to investigate why there are fails in Domain Access and Domain Config.

PR #216, opened 2016-04-27
allow users to manage other entities with domains

claire-remy:
This allows users to add their own entities to the $text variable, allowing them to manage them according to domains (as is done with user and node).

### DIFF
--- a/domain_access/domain_access.module
+++ b/domain_access/domain_access.module
@@ -384,6 +384,11 @@ function domain_access_confirm_fields($entity_type, $bundle) {
       'description' => 'Make this user an editor on all domains.',
     ]
   ];
+
+  //Allow users to manage other entities
+  $moduleHandler = Drupal::moduleHandler();
+  $moduleHandler->alter('domain_access_confirm_fields', $text);
+
   $id = $entity_type . '.' . $bundle . '.' . DOMAIN_ACCESS_FIELD;
 
   if (!$field = \Drupal::entityManager()->getStorage('field_config')->load($id)) {


agentrickard:
Really needs some sort of test. This is an interesting addition because by itself, it doesn't do anything.  But I get it.

claire-remy:
I had to add it in order to process my custom entity (which basically does what Nodes do) and thought this could come in handy for other users if they have custom entities to process differently according to a domain :)

zerolab:
I think this would come in handy for OG or [Group](https://www.drupal.org/project/group) integration.

agentrickard:
Right. My point was that those other entities will have to provide their own access controls, if desired. The main module won't be responsible for doing that.

domain_access/domain_access.module
```diff
@@ -384,6 +384,11 @@ function domain_access_confirm_fields($entity_type, $bundle) {
       'description' => 'Make this user an editor on all domains.',
     ]
   ];
+
+  //Allow users to manage other entities
+  $moduleHandler = Drupal::moduleHandler();
```
agentrickard, line 389: Calls to the Drupal helper class must be prefixed with a backslash:

```
$moduleHandler = \Drupal::moduleHandler();
```

agentrickard:
The test for this should try to add a field to a Vocabulary definition, using the provided test module.

zerolab:
This PR needs a few things:
- [ ] A `domain_access.api.php` file describing the new alter hook.
- [ ] Update documentation to flag the need for access control and handling creating new custom entity bundles (a la `domain_access_node_type_insert()`)
- [ ] Extend `domain_access_domain_insert()` to create relevant actions
- [ ] Test(s)
- [ ] Code style linting

I made a start on `domain_access.api.php`:

``` php
<?php

/**
 * @file
 * API documentation file for the Domain Access module.
 */

/**
 * Allow users to manage domain access for entities other than node/user.
 *
 * Note that it is the module maintainer's responsibility to provide their own
 * access control. See domain_access_node_access() for an example.
 *
 * @param array $text
 *   A key-value array that holds the required Domain Access field text.
 *   The key is the entity machine name (e.g. user).
 *
 * @see domain_access_confirm_fields()
 */
function hook_domain_access_confirm_fields_alter(&$text) {
  $text['my_entity'] = [
    'type' => 'content',
    'label' => 'Send to all affiliates',
    'description' => 'Make this "My Entity" available on all domains.',
  ];
}
```

domain_access/domain_access.module
```diff
@@ -384,6 +384,11 @@ function domain_access_confirm_fields($entity_type, $bundle) {
...
+  $moduleHandler->alter('domain_access_confirm_fields', $text);
```
agentrickard, line 390: Is an alter proper here? Why not an invokeAll()?

agentrickard:
This is outdated. I started another try in the 216-field branch.

Rather than using an alter, I made the function callable. Problem is, you need to create the field storage first, so the tests currently fail.

See DomainAccessEntityFieldTest

Issue #204, opened 2016-03-01
DomainConfigOverrider returns empty $overrides


artreaktor:
This is related to #196 
Background: I have four languages enabled (en,de,zh-hant,uk). For de.example.com I need Chinese and Ukrainian to be disabled. 
So, i have
domain.config.de_example_com.language.entity.uk.yml
and
domain.config.de_example_com.language.entity.zh-hant.yml
for that. Everything works well until I switch to the default (en) language:
The last-added language isn't overridden in language-switcher and translation interface.

This happens when $this->domain is empty in DomainConfigOverrider::loadOverrides()
Temporary solved with
`$this->domain = $this->domainNegotiator->getActiveDomain(TRUE);`
in DomainConfigOverrider::initiateContext() 

The domain negotiator doesn't return active domain under certain circumstances. Probably it is related to caching somehow. Where is the right point to fix this?

agentrickard:
Well, first I do not understand the names of the YML override files.

* `domain.config.de_example_com.language.entity.uk.yml` is malformed and should be `domain.config.de_example_com.uk.yml` and `domain.config.de_example_com.language.entity.zh-hant.yml` should be `domain.config.de_example_com.zh-hant.yml`.

If changing the names of the files doesn't solve the issue, then it's a matter of tracking through the call stack to see when the initiateContext() code is being called and fails. If we knew what the condition was, then we could figure an answer.

agentrickard:
You might also look at this method, which might be setting the wrong language context.

```
  /**
   * {@inheritdoc}
   */
  public function getCacheableMetadata($name) {
    $this->initiateContext();
    $metadata = new CacheableMetadata();
    if (!empty($this->domain)) {
      $metadata->addCacheContexts(['url.site', 'languages:language_interface']);
    }
    return $metadata;
  }
```

It's not clear to me which cache context we should use for languages.

agentrickard:
Fixed. 

   d460894..39947dd  8.x-1.x -> 8.x-1.x

PR #218, opened 2016-05-11
Linting and cleanup

zerolab:
This PR aims to fix a number of linting errors, remove unused use statements and cleanup.

Will look at replacing all deprecated usage next.

### DIFF
--- a/domain/domain.api.php
+++ b/domain/domain.api.php
@@ -12,10 +12,8 @@
  *
  * use Drupal\domain\DomainInterface;
  *
- *
  * @param array $domains
  *   An array of $domain record objects.
- *
  */
 function hook_domain_load(array $domains) {
   // Add a variable to the $domain.
@@ -55,8 +53,8 @@ function hook_domain_request_alter(\Drupal\domain\DomainInterface &$domain) {
  *
  * @return array
  *   An array of operations which uses a unique string key and requires the
- *  elements 'title' and 'url'; the 'query' value is optional, and used
- *  for link-actions with tokens
+ *   elements 'title' and 'url'; the 'query' value is optional, and used
+ *   for link-actions with tokens
  */
 function hook_domain_operations(\Drupal\domain\DomainInterface $domain) {
   // Add aliases to the list.
@@ -85,6 +83,7 @@ function hook_domain_operations(\Drupal\domain\DomainInterface $domain) {
  *   The HTTP_HOST string value being validated, such as one.example.com.
  *   Note that this is checked for uniqueness separately. This value is not
  *   modifiable.
+ *
  * @return
  *   No return value. Modify $error_list by reference. Return an empty array
  *   or NULL to validate this domain.
@@ -104,11 +103,11 @@ function hook_domain_validate_alter(&$error_list, $hostname) {
  * Note that this hook does not fire for users with the 'administer domains'
  * permission.
  *
- * @param $query
+ * @param \Drupal\Core\Entity\Query\QueryInterface $query
  *   An entity query prepared by DomainSelection::buildEntityQuery().
- * @param $account
+ * @param \Drupal\Core\Session\AccountInterface $account
  *   The account of the user viewing the reference list.
- * @param $contect array
+ * @param array $context
  *   A keyed array passing two items:
  *   - entity_type The type of entity (e.g. node, user) that requested the list.
  *   - bundle The entity subtype (e.g. 'article' or 'page').
@@ -118,7 +117,7 @@ function hook_domain_validate_alter(&$error_list, $hostname) {
  */
 function hook_domain_references_alter($query, $account, $context) {
   // Remove the default domain from non-admins when editing nodes.
-  if ($entity_type == 'node' && !$account->hasPermission('edit assigned domains')) {
+  if ($context['entity_type'] == 'node' && !$account->hasPermission('edit assigned domains')) {
     $default = \Drupal::service('domain.loader')->loadDefaultId();
     $query->condition('id', $default, '<>');
   }
--- a/domain/domain.api.php
+++ b/domain/domain.api.php
@@ -13,7 +13,7 @@
  * use Drupal\domain\DomainInterface;
  *
  *
- * @param array $domain
+ * @param array $domains
  *   An array of $domain record objects.
  *
  */
@@ -36,10 +36,10 @@ function hook_domain_load(array $domains) {
  * To issue a redirect, as in the case of Domain Alias, set a redirect
  * property to a valid response code (301 or 302).
  *
- * @param DomainInterface $domain
+ * @param \Drupal\domain\DomainInterface $domain
  *   A domain object defined by Drupal\domain\DomainInterface.
  */
-function hook_domain_request_alter(DomainInterface &$domain) {
+function hook_domain_request_alter(\Drupal\domain\DomainInterface &$domain) {
   // Add a special case to the example domain.
   if ($domain->getMatchType() == DOMAIN_MATCH_EXACT && $domain->id() == 'example_com') {
     // Do something here.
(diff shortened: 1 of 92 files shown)


domain/domain.api.php
```diff
@@ -36,10 +36,10 @@ function hook_domain_load(array $domains) {
  * To issue a redirect, as in the case of Domain Alias, set a redirect
  * property to a valid response code (301 or 302).
  *
- * @param DomainInterface $domain
+ * @param \Drupal\domain\DomainInterface $domain
  *   A domain object defined by Drupal\domain\DomainInterface.
  */
-function hook_domain_request_alter(DomainInterface &$domain) {
+function hook_domain_request_alter(\Drupal\domain\DomainInterface &$domain) {
```
agentrickard, line 40: Isn't it best practice for the module to declare a `use` here? Is this what core does?

domain/src/Access/DomainAccessCheck.php
```diff
@@ -10,7 +10,7 @@
 use Drupal\Core\Access\AccessCheckInterface;
 use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Session\AccountInterface;
-use Drupal\domain\DomainInterface;
+use Drupal\domain\Entity\Domain;
```
agentrickard: Doesn't using the Interface here make the class swappable?

domain/src/Controller/DomainControllerBase.php
```diff
@@ -41,9 +40,9 @@ class DomainControllerBase extends ControllerBase {
   /**
    * Constructs a new DomainControllerBase.
    *
-   * @param \Drupal\Core\Config\Entity\ConfigEntityStorage $entity_storage
+   * @param \Drupal\Core\Entity\EntityStorageInterface $entity_storage
    *   The storage controller.
-   * @param \Drupal\Core\Entity\EntityManagerInterface $entity_manager
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_manager
```
agentrickard, line 40: If you replaced all of these you are my new hero.

domain/src/DomainListBuilder.php
```diff
@@ -34,13 +34,10 @@ public function getFormId() {
    */
   public function getOperations(EntityInterface $entity) {
     $operations = parent::getOperations($entity);
-    $destination = drupal_get_destination();
+    $destination = \Drupal::destination()->getAsArray();
     $default = $entity->isDefault();
     $id = $entity->id();
-
-    // Get CSRF token service.
-    $token_generator = \Drupal::csrfToken();
```
agentrickard, line 42: This was a security measure. What is it replaced with?

domain/src/DomainLoaderInterface.php
```diff
@@ -30,7 +29,7 @@ public function load($id, $reset = FALSE);
   /**
    * Gets the default domain object.
    *
-   * @return Drupal\domain\DomainInterface | NULL
+   * @return \Drupal\domain\DomainInterface | NULL
```
agentrickard, line 32: Talking to Crell, we probably shouldn't return NULL. Is that coming from our code or from core?

domain/src/DomainValidator.php
```diff
@@ -104,7 +105,7 @@ public function validate(DomainInterface $domain) {
       }
     }
     // Allow modules to alter this behavior.
-    \Drupal::moduleHandler()->invokeAll('domain_validate', $error_list, $hostname);
+    \Drupal::moduleHandler()->invokeAll('domain_validate', array($error_list, $hostname));
```
agentrickard, line 108: Do we have a test for this yet? 

I will admit that I found the new hook syntax confusing.

domain_access/src/DomainAccessManager.php
```diff
@@ -97,6 +95,7 @@ public static function getDefaultValue(FieldableEntityInterface $entity, FieldDe
     switch ($entity->getEntityType()) {
       case 'user':
       case 'node':
+        // @TODO $this is no accessible in a static context.
```
agentrickard, line 93: I don't understand this comment.

agentrickard:
I'm in an airport, and added some comments for further (better) review.

You are my new favorite person.

domain/domain.api.php
zerolab, line 40: I am in two worlds about it. Core is inconsistent. e.g. see `hook_block_view_alter()`

domain/src/DomainListBuilder.php
zerolab, line 42: I've removed it as `$token_generator` is not used anywhere in `getOperations()`. I've check a number of other entity ListBuilders (e.g. `VocabularyListBuilder`) and none use a token.

domain/src/DomainValidator.php
zerolab, line 108: Good catch. We do not have a test for this yet. (note to self: would need to go in domain_test.module)

There is more to this though, `domain.api.php` defines an alter hook (`hook_domain_validate_alter()`) where as the like above is a hook. So it should probably be changed to `\Drupal::moduleHandler()->alter('domain_validate', $error_list, $hostname, $context);`

Which one do you prefer?

domain_access/src/DomainAccessManager.php
zerolab, line 93: Could be a linter hiccup, but from what I managed to find online, it has to do with the fact that `getDefaultValue()` is a static function and as per http://php.net/manual/en/language.oop5.static.php 

> Because static methods are callable without an instance of the object created, the pseudo-variable $this is not available inside the method declared as static.

I must admit I am not entirely sure what to do with it (if anything). Could use someone else's opinion.

domain/src/Access/DomainAccessCheck.php
zerolab: Good point. Tweaked as it was for type hinting in the `access()` method.
Somehow missed the fact that `getActiveDomain()` does return a DomainInterface

domain/src/DomainLoaderInterface.php
zerolab, line 32: This comes from us. See [loadDefaultDomain()](https://github.com/agentrickard/domain/blob/8.x-1.x/domain/src/DomainLoader.php#L71)

I did a quick search on the use of `loadDefaultDomain()` and it is mostly checking that it returns something. Bar `DomainSubscriber::onKernelRequestDomain()`

domain/src/DomainValidator.php
zerolab, line 108: On further thought, I believe we should do incremental PRs, rather than one big, does it all PR.
Thus fixing this hook call (depending on your preference on direct hook or alter hook) and proper tests should happen in its own little PR.

agentrickard, line 108: Makes sense to me.

domain_access/src/DomainAccessManager.php
agentrickard, line 93: No, I get it now. It's a static method because the Form system might call it at anytime, blindly.

We should refactor in a follow-up.

domain/src/DomainLoaderInterface.php
agentrickard, line 32: This is the default behavior of EntityInterface, so let's leave it alone.

domain/src/Access/DomainAccessCheck.php
```diff
@@ -1,16 +1,10 @@
 <?php
 
-/**
- * @file
- * Defines \Drupal\domain\Access\DomainAccessCheck.
- */
-
```
zerolab, line 7: Note that this is a fairly recent change: https://www.drupal.org/node/2304909

zerolab:
This is almost done. In my eager changes last night I missed a few things which makes more tests fail. Will follow up tonight.

Also, I think I found the reason the domain_config tests fail -- the language and language manager variables are not being initialised (https://github.com/agentrickard/domain/blob/8.x-1.x/domain_config/src/DomainConfigOverrider.php#L47-L56) and are not passed as arguments the [service](https://github.com/agentrickard/domain/blob/8.x-1.x/domain_config/domain_config.services.yml)

agentrickard:
Let's leave the config tests for a follow-up. There is already an issue for that #204.

zerolab:
I missed #204 somehow. Will update the PR tonight when it is ready for the final review.

zerolab:
OK, this is now ready for review.

Several notes:
- Only replaced a deprecated function call in `domain.drush.inc`
- The `domainAliasCreateTestAlias()` exception is fixed in https://github.com/agentrickard/domain/commit/e2d6571873deb2ff137710110d1706ac9aa4c774 from #215
- DomainConfig tests are discussed in #204, so not touched here.
- Created #222 for the `domain_validate` hook and tests
- Created #223 for the `DomainAccessManager::getDefaultValue()` static context issue

agentrickard:
I just did a local branch merge and am running tests there.

PR #230, opened 2016-06-12
Fixes up tests for default domain values.

agentrickard:


domain_access/src/DomainAccessManager.php
```diff
@@ -87,12 +87,11 @@ public function checkEntityAccess(EntityInterface $entity, AccountInterface $acc
...
   public static function getDefaultValue(FieldableEntityInterface $entity, FieldDefinitionInterface $definition) {
     $item = array();
-    switch ($entity->getEntityType()) {
+    switch ($entity->getEntityType()->id()) {
       case 'user':
       case 'node':
-        // @TODO $this is no accessible in a static context.
         /** @var \Drupal\domain\DomainInterface $active */
-        if ($active = $this->negotiator->getActiveDomain()) {
+        if ($active = \Drupal::service('domain.negotiator')->getActiveDomain()) {
```
zerolab: Cursory look, but shouldn't this service be injected?
It is fresh on my mind as I just read https://www.drupal.org/documentation/modules/serialization#comment-10989879

zerolab, line 94: Deleted previous comment. Looong week.
It obviously replaces the inject service as this is a static function.

PR #233, opened 2016-06-12
Test for field creation.

agentrickard:
New version of #216


agentrickard:
The test pass here is a false positive, because of the travis.yml configuration.

agentrickard:
So the hold-up here is that other entities have to create their own field storage before we can add the field. I'm not sure I feel compelled to provide a helper function for that. (Currently stubbed as `domain_access_confirm_storage`.)

I may just put it in the test module as an example. That should work and make the tests pass.

Issue #234, opened 2016-06-13
Fix test failures for localhost


agentrickard:
Forked from https://github.com/agentrickard/domain/pull/232#issuecomment-225463114

Looks like we have some inconsistencies in the test runner when localhost is the primary domain instead of example.com.

agentrickard:
This should be accounted for in domainCreateTestDomains() but is worth checking.

PR #235, opened 2016-06-17
[wip] Fix Travis test failures

zerolab:
Related #234
- Changed the Travis runner to use example.com
- Removed Drupal 8.0
- Fixed machine name generation logic in tests to follow that of ConfigBase.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,4 +76,8 @@ before_script:
   - until curl -s example.com:8080; do true; done > /dev/null
 
 script:
-  - php core/scripts/run-tests.sh --verbose --color --concurrency 4 --php `which php` --url http://example.com:8080 "domain_access,domain,domain_alias,domain_config,domain_source"
+  - php core/scripts/run-tests.sh --verbose --color --concurrency 4 --php `which php` --url http://example.com:8080 domain
+  - php core/scripts/run-tests.sh --verbose --color --concurrency 4 --php `which php` --url http://example.com:8080 domain_access
+  - php core/scripts/run-tests.sh --verbose --color --concurrency 4 --php `which php` --url http://example.com:8080 domain_alias
+  - php core/scripts/run-tests.sh --verbose --color --concurrency 4 --php `which php` --url http://example.com:8080 domain_config
+  - php core/scripts/run-tests.sh --verbose --color --concurrency 4 --php `which php` --url http://example.com:8080 domain_source
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,18 +10,18 @@ matrix:
 
 addons:
   hosts:
-    - one.localhost
-    - two.localhost
-    - three.localhost
-    - four.localhost
-    - five.localhost
+    - example.com
+    - one.example.com
+    - two.example.com
+    - three.example.com
+    - four.example.com
+    - five.example.com
 
 cache:
   directories:
     - $HOME/.composer/cache/files
 
 env:
-  - DRUPAL=8.0.x
   - DRUPAL=8.1.x
   - DRUPAL=8.2.x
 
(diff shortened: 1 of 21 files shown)


agentrickard:
I'd like to not make 'example.com' required. We should be able to write tests so that the root domain can vary, but the machine name is consistent.

What I'd like to know is which tests are failing, and is it because we're looking for the wrong text?  Perhaps we just need to ensure consistent labels as well as consistent machine_names.

agentrickard:
The ultimate goal is for the tests to work regardless of root domain, so long as wildcard DNS is used.

zerolab:
I definitely agree with making the tests work regardless of the root domain. Tried fixing this over the weekend

See 646296a, f4c567b and dac1da4

Most telling is dac1da4. When I run the tests for domain_access separately, they all come out green. When I run then after domain, there are failures.

domain_config is the difficult one to test with any domain as the test module adds domain-specific configuration and depends a lot on the machine name. To add to the complication, the generated domain machine name takes in consideration the port too. So localhost and localhost:8080 are correctly treated as two different machine names. I think the test could be changed to provide custom machine names like domain_one, domain_two and change the DomainConfig tests to generate those domains, rather than use `$hostname = $machine_name = $list[$i] . '.' . $base_hostname;`. 

Would that work?

agentrickard:
We may be misusing machine_names in some cases. They are supposed to be unique keys and should not be dependent on the inbound URL.

Right now they are all coded to be `$list[$i] . '.example.com'` in order for the Domain Config files to be recognized. Domain Config itself uses the machine name (the id() method) to load the file.

The generated machine name shouldn't use the port at all, because of the empty check:

```
  public function createMachineName($hostname = NULL) {
    if (empty($hostname)) {
      $hostname = $this->createHostname();
    }
    return preg_replace('/[^a-z0-9_]+/', '_', $hostname);
  }
```

I don't quite know why the Travis test runner wants the port 8080, but it shouldn't be relevant to the machine names.

I am very confused as to why the order of tests matters. That suggests that the Travis build isn't cleaning up after itself properly.

Do we need to split the test runner line into one call per test group?

agentrickard:
I'm going to run locally using `localhost, one.localhost` and see what happens.

Also note that I'm running via the UI

zerolab:
I get the test failures locally, so it is not only Travis.

There are 2 reasons we run the tests on port 8080 on Travis:
1. Running the built in PHP server on port 80 requires sudo, which in term spins up the old Travis VMs which are being phased out.
2. Both locally and on Travis I had permissions issues with running as sudo

Saying that, I pushed my WIP which now includes running the tests separately, which should finally pass.

zerolab:
As for `/[^a-z0-9_]+/`, it definitely filters out anything non-alphanumeric, but example.com:8080 becomes example_com_8080 as a machine name.

zerolab:
Much better!

Not sure why https://travis-ci.org/agentrickard/domain/builds/139057046 failed as the same test in my branch passes - https://travis-ci.org/zerolab/domain/builds/139057033. Maybe restart the jobs?

agentrickard:
Restarted those.

It's interesting, when I ran the tests locally against localhost/one.localhost/etc I got two test failures in the DomainAccessPermissions test, both, I think, related to entity loading.

agentrickard:
Hmm. One fail on re-test, PHP 7 Drupal 8.2 only.

```
Fail      Other      DomainValidatorTe   33 Drupal\domain\Tests\DomainValidator

    Server test for http://example.com:8080/ passed.
```

agentrickard:
Reran it and got a different fail?

```
Drupal\domain\Tests\DomainValidatorTest                        6 passes   1 fails                            

- Found database prefix 'simpletest820405' for test ID 9.
```

agentrickard:
I'm going to commit this, then retest locally and see what (if anything needs changing.)

Issue #115, opened 2016-01-19
Ensure unique numeric ids for use with node access


agentrickard:
When creating a domain record, we use machine names as the primary key.

However, the node access system still requires a numeric id. Currently, we generate that from code in `Drupal\domain\DomainCreator::createNextId()`.

This routine needs to be checked so that we ensure:

* IDs are never repeated
* Test cases prove the above

That probably means storing a simple state variable, which might also need to be added to `domain.settings.yml`. A simple name like `domain_last_id` should work for this purpose.

anishsheela:
Working on it.

agentrickard:
We can add the test case to `Drupal\domain\Tests\DomainCreateTest` or spin up a `DomainUniqueIdTest` class.

agentrickard:
I was looking through this today and I think we have to implement some form of "ok, you're deleting a domain, would you like to reassign users/content."

Shouldn't be too hard, given that we know what data to update.

zerolab:
Is storing the last domain id as a state variable the best option? https://www.drupal.org/developing/api/8/state states [sic]

> You can reset a system, losing all state. Its configuration remains.

Having seen how even the D6/7 variables could fail dramatically on large site, I may be a bit overly cautious.

Your idea of reassigning the content on domain deletion is very good. If there are only 2 domains, should this step be skipped? i.e. all content from the deleted domain is automagically re-assigned to the remaining one.

agentrickard:
State variable is, I think, the last resort, as State is not exportable and domains are. And the problem is that domain_id must be an integer (due to the {node_access} table), and is part of the export. So it becomes very easy to repeat ids of deleted domains.

In an ideal world, we could convert the machine_name to an integer value, perhaps using http://php.net/manual/en/function.crc32.php

As for reassigning content, yes, if only two domains, the assignment would not allow the user a choice (though it still needs to happen). We have this feature in the D7 version, I'm just giving it a higher priority.

agentrickard:
Sadly, crc32 produces integers too large to store in the current table, and I don't want to push that to BIGINT.

zerolab:
Hi Ken,

Had a quick look at the schema, and unless I am missing something glaringly obcious, it declares domain_id as an integer. MAX_INT_SIZE is at least 32-bit. This should hold the crc32 result, without the need to bump it to BIGINT.

Also, since the node_access table uses `int(10)`, i.e. unsigned int, we'd need to do something like:
```php
[...]

PR #238, opened 2016-06-26
Create unique numeric ids for domains.

agentrickard:


zerolab:
`DomainCreateTest` needs updating. Line 37 still checks for the old style ids.
`$this->assertTrue($new_domain->getDomainId() == 1, 'Domain id set properly.');`

What is the best way to handle the change on site that already use this on production? Would an update hook that flags the need for node access rebuild suffice?

(Making note that this fixes #115)

agentrickard:
Re-indexing node_access would not help. The issue is that new domains would have the domain_id created in a different way and therefore be non-sequential. 

That actually should affect current sites, and those "old" ids could be considered stable. 

We are not regenerating ids for existing domains. You could do that in an update hook and then rebuild node_access, but I don't think it is necessary.

(Also note that this PR is flawed, and we need to be generating off machine_name/id not hostname, and we can't do it in preSave, because that introduces volatility should the machine name change.)

agentrickard:
@zerolab This test fail may have to do with the 8080 port/path in that one test.

domain/src/Entity/Domain.php
```diff
@@ -173,10 +173,9 @@ public static function preCreate(EntityStorageInterface $storage_controller, arr
       'status' => 1,
       'weight' => count($domains) + 1,
       'is_default' => (int) empty($default),
-      // {node_access} still requires a numeric id.
```
zerolab, line 176: The `$creator` variable above (`$creator = \Drupal::service('domain.creator');`) is unused now.

domain/src/Entity/Domain.php
```diff
@@ -349,6 +348,21 @@ public function preSave(EntityStorageInterface $storage_controller) {
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function createDomainId() {
+    // We cannot reliably use sequences (1, 2, 3) because those can be different
+    // across environments. Instead, we use the crc32 hash function to create a
+    // unique numeric id for each domain.
+    $id = preg_replace('/[^a-z0-9_]+/', '_', $this->getHostname());
+    $this->domain_id = (int) crc32($id);
```
zerolab: Should this not use the `createMachineName` method from the injected `DomainCreator` service?

zerolab:
> (Also note that this PR is flawed, and we need to be generating off machine_name/id not hostname, and we can't do it in preSave, because that introduces volatility should the machine name change.)

Is this an issue though? In many ways we could treat this as special entity id in that we only generate it once, during the creation based on the hostname and the creation timestamp. This doesn't change on consequent hostname/machine name changes. And even if the machine name changes in a `hook_entity_insert()` it should not matter.

domain/src/Entity/Domain.php
agentrickard: If the service is available, then yes, we should fix this instance and the one above so that we have a consistent method call.

That means we can move createDomainId() to the creator service.

agentrickard:
I don't think the machine-name / hostname issue matters. The bigger issue was only creating on insert, which I did fix.

Hostname already has to be unique, so we don't need timestamp.

agentrickard:
I removed the $creator call and simplified createDomainId() to use the hostname, which makes the method self-contained.

zerolab:
👍 and the tests pass too.

PR #243, opened 2016-06-30
Issue #242: remove obsolete patch instructions

### LINKED ISSUE #242: Remove obsolete patch instructions for Domain Config
Currently the Domain Config module's README file lists as a dependency:

* The patch from [#2662196 Cache route by Uri and not just Query+Path](https://www.drupal.org/node/2662196)

It looks like this core patch is no longer needed given the override to the ```router.route_provider``` class.

nedjo:
Fixes #242.

### DIFF
--- a/domain_config/README.md
+++ b/domain_config/README.md
@@ -61,4 +61,3 @@ Dependencies
 ============
 
 - Domain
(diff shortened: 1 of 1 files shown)


zerolab:
LGTM

Issue #244, opened 2016-07-02
Merge in partial config overrides instead?


nedjo:
Currently config overrides replace the full item. This can lead to unexpected or undesired outcomes.

Example:
* On domain B, register an override of ```system.site``` to customize the name per domain.
* Later, edit the original ```system.site.page.404``` value. The edit is not carried through to Domain B, since, to override the name, all ```system.site``` values were overridden.

An alternative would be to adopt the approach used in the [Config Override](https://www.drupal.org/project/config_override) module, where partial overrides are merged in. See [ModuleConfigOverrides](http://cgit.drupalcode.org/config_override/tree/src/ModuleConfigOverrides.php).

agentrickard:
Can you export partial config yml files like that? Won't they break on import because of missing schema?

agentrickard:
Oh, I see what that module is doing. It seems to be bypassing the core import/export model, which is what we're using now.

That would be a paradigm shift in how we think about config.

Another option: I wonder if we could use a placeholder definition in the yml output to indicate "inherit from the default config"?

Something like so:

```
name: 'Four'
mail: admin@example.com
slogan: '*INHERIT*'
page:
  403: '*INHERIT*'
  404: '*INHERIT*'
  front: /node/2
admin_compact_mode: false
weight_select_max: 100
langcode: en
```

agentrickard:
As far as I can tell, this Just Works (tm).

domain.config.four_example_com.system.site.yml
```
name: 'Four'
page:
  front: /node/2
```

See #245.

If that's the case, then this becomes a documentation issue.

nedjo:
Nice!

agentrickard:
We still need to verify that Drupal thinks this is valid config. I suspect that Config Inspector may show that core thinks the schema is invalid because it is incomplete.

nedjo:
Ah, this works because [ConfigFactory::loadOverrides()](https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Config!ConfigFactory.php/function/ConfigFactory%3A%3AloadOverrides/8.2.x) calls [NestedArray::mergeDeepArray()](https://api.drupal.org/api/drupal/core!lib!Drupal!Component!Utility!NestedArray.php/function/NestedArray%3A%3AmergeDeepArray/8.2.x), so we only need to set changes.

But actually I think we're fine for config schema. What I assume is happening:
* When we create an override, it's saved as simple configuration, regardless of the original configuration type being overridden.
[...]

PR #241, opened 2016-06-30
Document how to override configuration by type

nedjo:
Fixes #240.

PR #245, opened 2016-07-04
Remove some configuration options for testing.

agentrickard:
This seems to indicate that partial configuration exports are allowable.

Could use validation in the real world.

See #244.


zerolab:
Do we need this PR? I think the important bits are covered in #241

agentrickard:
Yea. It changes the test and proves the feature.

zerolab:
Good point.

Worth restarting the tests, to ensure it not the occasional failures we see on Travis.

Issue #239, opened 2016-06-27
Domain Access fields: Improve UI for node form 


bjaxelsen:
The node edit form places advanced fields (promote to front, sticky, menu settings etc.) in an accordion to keep the basic content editing simple and usable. I think we should do the same for the Domain Access fields.

Also, we might want to provide the current domain as default if no default is given. To some users the concept of controlling domain access can be somewhat complicated, so if we can turn this into something that most users don't need to take care of, it would be better.

How about adding this form alter in domain_access.module?

```
/**
 * Implements hook_form_BASE_FORM_ID_alter().
 * 
 * Set a domain for domain access by default and move Domain-specific fields to
 * an advanced tab
 */
function domain_access_form_node_form_alter(&$form, FormStateInterface $form_state, $form_id) {
  if (isset($form['field_domain_access'])) {
    if(empty($form['field_domain_access']['widget']['#default_value'])) {
      // Get current domain
      $domain = \Drupal::service('domain.negotiator')->getActiveDomain()->id();
      if (isset($form['field_domain_access']['widget']['#options'][$domain])) {
        $form['field_domain_access']['widget']['#default_value'] = [$domain];
      }
    }
    // Move to the tabs on the node form
    $form['field_domain_access']['#group'] = 'domain';
    if (isset($form['field_domain_all_affiliates'])) {
      $form['field_domain_all_affiliates']['#group'] = 'domain';
    }
    $form['domain'] = [
      '#type' => 'details',
      '#title' => t('Domain'),
      '#group' => 'advanced',
      '#attributes' => [
        'class' => ['node-form-options']
      ],
      '#attached' => [
        'library' => ['node/drupal.node'],
      ],
      '#weight' => 100,
      '#optional' => TRUE
    ];
  }
}
```

agentrickard:
That default to current domain should already be set. See testDomainAccessDefaultValue().

Putting it in the sidebar would, to me, be a configuration option, which is what we did in D7. 

Can we make that part conditional on a setting, but default to TRUE?

agentrickard:
Also wondering: how does this play with modules like Field Group?

PR #249, opened 2016-07-18
Move domain access fields on node forms

bjaxelsen:
I've made this based on #239. It allows us to configure Domain Access to move the Domain Access fields to an advanced tab on the node forms. This is good for editors who may not understand the meaning of the fields - they can concentrate on the main content fields.

I don't think this interferes in a bad way with the field group module. The advanced groups on the node edit forms is already available and by making this a configurable option it can easily be turned off in case you want to move the fields to some other field group etc.


domain_access/config/install/domain_access.settings.yml
```diff
@@ -0,0 +1 @@
+node_advanced_tab: 'TRUE'
```
zerolab: This needs to be simply `true`.

The tests fail with:

```
Drupal\Core\Config\Schema\SchemaIncompleteException: Schema errors for
    domain_access.settings with the following errors:
    domain_access.settings:node_advanced_tab variable type is string but
    applied schema class is Drupal\Core\TypedData\Plugin\DataType\BooleanData
    in Drupal\Core\Config\Testing\ConfigSchemaChecker-&gt;onConfigSave()
```

zerolab:
Checked this on my local test build. Does what it says on the tin. When the setting is enabled, the fields will move to the advanced groups even if they are added to a field group in the form display settings. The only downside is that the field group then appears as empty, with just the title.

I think it would be helpful to add a note on the domain_access settings page about that to avoid any confusion.

bjaxelsen:
Yes, I think it makes sense to only move the fields if they are not put in a field group already. I've changed this and modified the help text on the configuration form.

zerolab:
I am not entirely sure about this. The core fields that are in the advanced tab do not react to them being in a field group. So maybe this should do the same.

domain_access/domain_access.routing.yml
```diff
@@ -0,0 +1,7 @@
+domain_config.admin:
```
agentrickard, line 1: This route is for domain_access, not domain_config.

domain/domain.links.task.yml
```diff
@@ -1,5 +1,6 @@
-entity.domain.edit_form:
```
agentrickard, line 1: I don't understand why this task is being removed here.

agentrickard:
OK, I've moved this into the general settings branch that I'm working on, Thanks!

domain/domain.links.task.yml
bjaxelsen: sorry, think I made a mistake here

agentrickard: It doesn't seem to affect anything. See the new PR in https://github.com/agentrickard/domain/pull/264

Issue #250, opened 2016-07-18
field_domain_access should not be requried for entity type user


nedjo:
When Domain Access is enabled, a user with suitable permissions sees a required field, ```field_domain_access```, when editing any user record. Often, an editor may wish to edit a user account (for example, to approve it) without assigning any domain access. However, because this field is required, it's not possible to edit a user record without assigning the user editor access to at least one domain.

Proposed resolution: don't make ```field_domain_access``` required for entity type user.

agentrickard:
Seems legit.

zerolab:
This can be closed now that #252 is merged.

PR #252, opened 2016-07-18
Don't require field_domain_access on entity type 'user'

nedjo:
Issue: #250.


agentrickard:
That failure is acceptable.

Issue #253, opened 2016-07-28
Unable to get alias working


mickaelperrin:
Hi,

Thanks for working on the port to Drupal 8 of this module. I fighted with it a good part of the day, got config override working but only with direct match. I deep dive in the code and there are some parts I don't undestand. Could you give me some infos ?

## What I want to achieve

I intend to deploy 4 websites on the same Drupal. They will share some common entity types but will have for example specific users and themes.

So far, I tried the config with two domain and the following synonyms domain (depending on the configuration / development stage):

- www.domain.fr : www.domain.fr.docker, devel.domain.fr.docker, debug.domain.fr.docker
- extranet.domain.fr : extranet.domain.fr.docker, devel.extranet.domain.fr, debug.extranet.domain.fr

## What I have done so far 

Create two domains : 

- extranet.domain.fr with the following aliases extranet.domain.fr.docker, devel.extranet.domain.fr, debug.extranet.domain.fr
- www.domain.fr with the following aliases www.domain.fr.docker, devel.domain.fr.docker, debug.domain.fr.docker, and set as **default**

## What I don't understand

When accessing the domain `extranet.domain.fr.docker`, the `setRequestDomain` method is called and the `$domain` is set to the default domain before the `domain_request_allter` hook is called.

Is it intended ?

Do you have some tips to achieve what I want ?

mickaelperrin:
Just added the PR #254, to explicit what I did to resolve this issue.

agentrickard:
This does indeed look like the correct solution.

agentrickard:
Fixed in PR #255.  Nice work!

PR #254, opened 2016-07-29
[BUGFIX #253] Aliasing not working if non default

mickaelperrin:
- PB: aliases defined for a non default domain are not used as the
  hook is called with the default domain.
- FIX: only switch to the default domain if the hook with the
  requested domain returns nothing

### DIFF
--- a/domain/src/DomainNegotiator.php
+++ b/domain/src/DomainNegotiator.php
@@ -70,10 +70,6 @@ class DomainNegotiator implements DomainNegotiatorInterface {
       // If the load worked, set an exact match flag for the hook.
       $domain->setMatchType(DOMAIN_MATCH_EXACT);
     }
-    // Fallback to default domain if no match.
-    elseif ($domain = $this->domainLoader->loadDefaultDomain()) {
-      $domain->setMatchType(DOMAIN_MATCH_NONE);
-    }
     // If a straight load fails, create a base domain for checking. This data
     // is required for hook_domain_request_alter().
     else {
@@ -90,6 +86,16 @@ class DomainNegotiator implements DomainNegotiatorInterface {
     $id = $domain->id();
     if (!empty($id)) {
       $this->setActiveDomain($domain);
+    } else {
+      // Fallback to default domain if no match.
+      if ($domain = $this->domainLoader->loadDefaultDomain()) {
+        $this->moduleHandler->alter('domain_request', $domain);
+        $domain->setMatchType(DOMAIN_MATCH_NONE);
+        $id = $domain->id();
+        if (!empty($id)) {
+          $this->setActiveDomain($domain);
+        }
+      }
     }
   }
 
--- a/domain/src/DomainNegotiator.php
+++ b/domain/src/DomainNegotiator.php
@@ -86,15 +86,14 @@ class DomainNegotiator implements DomainNegotiatorInterface {
     $id = $domain->id();
     if (!empty($id)) {
       $this->setActiveDomain($domain);
-    } else {
-      // Fallback to default domain if no match.
-      if ($domain = $this->domainLoader->loadDefaultDomain()) {
-        $this->moduleHandler->alter('domain_request', $domain);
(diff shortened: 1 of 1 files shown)


domain/src/DomainNegotiator.php
```diff
@@ -90,6 +86,16 @@ public function setRequestDomain($httpHost, $reset = FALSE) {
     $id = $domain->id();
     if (!empty($id)) {
       $this->setActiveDomain($domain);
+    } else {
+      // Fallback to default domain if no match.
+      if ($domain = $this->domainLoader->loadDefaultDomain()) {
```
zerolab: use `elseif`

nitpick on codestyle:

``` php
}
else {
```

instead of `} else {`

domain/src/DomainNegotiator.php
```diff
@@ -91,6 +87,15 @@ public function setRequestDomain($httpHost, $reset = FALSE) {
     if (!empty($id)) {
       $this->setActiveDomain($domain);
     }
+    // Fallback to default domain if no match.
+    elseif ($domain = $this->domainLoader->loadDefaultDomain()) {
+      $this->moduleHandler->alter('domain_request', $domain);
+      $domain->setMatchType(DOMAIN_MATCH_NONE);
+      $id = $domain->id();
+      if (!empty($id)) {
```
zerolab, line 95: Maybe `if (!empty($domain->id()))` ?
We are just using `$id` for the if.

_edit_ I see the same is used above

mickaelperrin, line 95: Thanks for taking time to review the PR. Indeed I didn't want to change too much the original code.

I was mainly wondering if the approach taken to resolve my issue was the good one, or if I was missing something on how `domain` works.

Edit: If the PR is confirmed as a bug resolution, I will improve the code as you requested.

zerolab:
@mickaelperrin would be great to have a test for this.
Will definitely help validate the behaviour.

mickaelperrin:
ok will try to look at this. Currently work with phpunit and behat, I didn't have tried simpletest so far. If you have any good ressources that would help me to dive into this faster, you're welcome.

zerolab:
https://www.drupal.org/simpletest is a good start.
But do have a look at the existing tests to get an idea of what is needed.

agentrickard:
For a test, we can set both test domains in an array and then loop through lines 50-70 of DomainAliasNegotiatorTest.php.

The reason the tests don't catch this now is that the first domain ($alias_domain) is the default.

agentrickard:
```
    // Test the response of the default home page.
    /** @var \Drupal\domain\Entity\Domain $domain */
    foreach (\Drupal::service('domain.loader')->loadMultiple() as $domain) {
      $alias_domains[] = $domain;
      $this->drupalGet($domain->getPath());
      $this->assertRaw($domain->label(), 'Loaded the proper domain.');
      $this->assertRaw('Exact match', 'Direct domain match.');
    }

    // Now, test an alias for each domain.
    foreach ($alias_domains as $alias_domain) {
      $this->domainAliasCreateTestAlias($alias_domain);
      $pattern = '*.' . $alias_domain->getHostname();
      $alias = \Drupal::service('domain_alias.loader')->loadByPattern($pattern);
      $alias_domain->set('hostname', 'two.' . $alias_domain->getHostname());
      $alias_domain->setPath();
      $url = $alias_domain->getPath();
      $this->drupalGet($url);
      $this->assertRaw($alias_domain->label(), 'Loaded the proper domain.');
      $this->assertRaw('ALIAS:', 'No direct domain match.');
      $this->assertRaw($alias->getPattern(), 'Alias match.');

      // Test redirections.
      // @TODO: This could be much more elegant: the redirects break assertRaw()
      $alias->set('redirect', 301);
      $alias->save();
      $this->drupalGet($url);
      $alias->set('redirect', 302);
      $alias->save();
      $this->drupalGet($url);
    }
```

I'll test this in a bit.

agentrickard:
Moved that to #255.

mickaelperrin:
Thanks for having merged this and updated tests.

PR #255, opened 2016-07-31
Add test for non-default domain aliases.

agentrickard:
This is a spinoff of #254 with tests that show the problem.


agentrickard:
I've added in the code from #254, and the tests are still failing. So we need to dig a bit and see why.

PR #257, opened 2016-08-08
256 default

agentrickard:
Corrects the default values issue. The test is not complete, as we need to figure our how to click the submit button in the second test rather than use drupalPostForm().

### DIFF
--- a/domain_access/src/Tests/DomainAccessDefaultValueTest.php
+++ b/domain_access/src/Tests/DomainAccessDefaultValueTest.php
@@ -1,0 +1,102 @@
+<?php
+
+namespace Drupal\domain_access\Tests;
+
+use Drupal\Component\Render\FormattableMarkup;
+use Drupal\Core\Database\Database;
+use Drupal\domain\Tests\DomainTestBase;
+
+/**
+ * Tests the domain access handling of default field values.
+ *
+ * @see https://www.drupal.org/node/2779133
+ *
+ * @group domain_access
+ */
+class DomainAccessDefaultValueTest extends DomainTestBase {
+
+  /**
+   * Modules to enable.
+   *
+   * @var array
+   */
+  public static $modules = array('domain', 'domain_access', 'field', 'field_ui');
+
+  /**
+   * Test the usage of DomainAccessManager::getDefaultValue().
+   */
+  public function testDomainAccessDefaultValue() {
+    $this->admin_user = $this->drupalCreateUser(array(
+      'bypass node access',
+      'administer content types',
+      'administer node fields',
+      'administer node display',
+      'administer domains',
+      'publish to any domain',
+    ));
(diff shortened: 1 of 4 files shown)


zerolab:
The test failures are possibly related to https://github.com/travis-ci/travis-ci/issues/2250.

On the test itself, `drupalPostForm()` does take a `$submit` parameter:

> $submit: Value of the submit button whose click is to be emulated. For example, t('Save'). The processing of the request depends on this value. For example, a form may have one button with the value t('Save') and another button with the value t('Delete'), and execute different code depending on which one is clicked.

I think that is the closest we get before needing to do the whole behat/BrowserTestBase thing which is much much slower if I read things correctly.

agentrickard:
The tests are running fine locally.

The $submit value isn't what I'm looking for, as drupalPostForm() is a programmatic form submission -- you have to populate $edit with form values, so that doesn't mimic this experience. I want to actually do a GET /node/1/edit and click the Save button.

There is a way to do that, but I forget what it is.

agentrickard:
I was thinking of clickLink(), but there is apparently no corresponding version for buttons.

I think, then, that the test we have now is accurate:

```
    // Now save the node with the values set.
    $edit = [
      'title[0][value]' => 'Test node update',
    ];
    $this->drupalPostForm('node/1/edit', $edit, 'Save');
```

This changes the title but doesn't touch the domain access values, which are preserved.

(Note: the behavior also works if I just test manually.)

I think my related question is "Should we just abstract this to all entity types just in case?"

zerolab:
> "Should we just abstract this to all entity types just in case?"

May seem like overkill, but it'd be a good idea, imho.

offtopic: will look at submitting a PR with Travis tweaks in the coming day.

agentrickard:
It's easy code, just use hook_form_alter() and check for an entity, then pipe it to the helper function domain_access_default_form_values().

domain_access/src/Tests/DomainAccessDefaultValueTest.php
```diff
@@ -0,0 +1,102 @@
...
+    $this->drupalPostForm('node/1/edit', $edit, 'Save');
+
+    // Load the node.
+    $node = \Drupal::entityTypeManager()->getStorage('node')->load(1);
+    $this->assertTrue($node, 'Article node created.');
+    // Check that the values are set.
+    $values = \Drupal::service('domain_access.manager')->getAccessValues($node);
+    $this->assertTrue(count($values) == 1, 'Node saved with one domain record.');
+    $allValue = \Drupal::service('domain_access.manager')->getAllValue($node);
+    debug($allValue);
```
agentrickard: Remove the debug.

This test will be better if we resave the node with multiple domains (as the admin user) and change the all affiliates value. Then we should save as the test user and ensure that the values are retained.

agentrickard:
Added support for all entity types.

Issue #137, opened 2016-01-19
Tests for all module hooks


agentrickard:
Once domain.api.php is finished, ensure we have proper test coverage in the domain_test module for each hook.

andypost:
at least it makes sense to start tracking

PR #260, opened 2016-08-15
two hooks had been added

wprocessor:


domain/src/Tests/DomainHooksTest.php
```diff
@@ -21,7 +21,7 @@ class DomainHooksTest extends DomainTestBase {
   /**
    * Tests domain loading.
    */
-  public function testHookDomainLoad() {
+  public function __testHookDomainLoad() {
```
zerolab: the test method names should not start with underscores.
This is likely the reason the tests fail with:

```
---- Drupal\domain\Tests\DomainHooksTest ----
Status    Group      Filename          Line Function                            
--------------------------------------------------------------------------------
Fail      Requiremen TestBase.php      1014 Drupal\simpletest\TestBase::run()  
    No test methods found.
```

andypost:
This about issue #137

wprocessor:
Please, use latest version to test, I removed underscores from function (I used underscores to disable some test to save time to running tests)

domain/src/Tests/DomainHooksTest.php
```diff
@@ -2,7 +2,7 @@
 
 namespace Drupal\domain\Tests;
 use Drupal\Component\Render\FormattableMarkup;
-
+use Drupal\domain\DomainListBuilder;
```
agentrickard, line 5: This doesn't seem to be used in the tests below.

domain/tests/modules/domain_test/domain_test.module
```diff
@@ -13,3 +13,39 @@ function domain_test_domain_load(array $domains) {
...
+  $domain->addProperty('foo1', 'bar1');
+}
+
+/**
+ * Implements hook_domain_operations
+ */
+function domain_test_domain_operations(\Drupal\domain\DomainInterface $domain) {
+  // Add aliases to the list.
+  $domain->addProperty('foo2', 'bar2');
+  return $operations;
```
agentrickard, line 30: This isn't really a proper test of what this hook does.

domain/tests/modules/domain_test/domain_test.module
```diff
@@ -13,3 +13,39 @@ function domain_test_domain_load(array $domains) {
...
+  }
+}
+
+/**
+ * Implements hook_domain_references_alter
+ */
+function domain_test_domain_references_alter($query, $account, $context) {
+  if ($context['entity_type'] == 'node' && !$account->hasPermission('edit assigned domains')) {
+    $default = \Drupal::service('domain.loader')->loadDefaultId();
+    $query->condition('id', $default, '<>');
```
agentrickard, line 49: This test would then need to load a node form and check that the options are as expected.

agentrickard:
I missed that this also fixed #358, so I'm going to refactor these tests.

PR #261, opened 2016-08-15
defines are not loaded in all contexts

kdebisschop:
We are getting this error in our PHP error logs:

> PHP Notice:  Use of undefined constant DOMAIN_MATCH_EXACT - assumed 'DOMAIN_MATCH_EXACT' in /mnt/www/html/fensitestg/docroot/modules/domain/domain/src/DomainNegotiator.php on line 72

In some requests, the defined constants in **domain.module** are not available under Drupal 8. I offer a single unified constants file as one way to ensure the constant is there across the various modules that are part of the **domain** repository. If you think this is not the best way to address the issue, I'd be glad to hear your thoughts and submit an alternate pull request.

The most obvious alternative to me was to create separate constant files for **domain** and **domain_access**. But since **domain_content** also relies on the `DOMAIN_ACCESS_*` constants, it seemed sensible to me to rely on the fact that the **domain_**\* modules depend on the **domain** modules and just out them all in one file.


domain/src/Constants.php
```diff
@@ -0,0 +1,37 @@
+<?php
+/**
+ *
+ *
+ * PHP Version 5
+ *
+ * @author Karl DeBisschop <karl.debisschop@fen.com>
```
zerolab: This docblock does not correspond to d.o. coding standards

domain/src/Constants.php
```diff
@@ -0,0 +1,37 @@
...
+
+class Constants {
+  /**
+   * Defines record matching types when dealing with request alteration.
+   *
+   * @see hook_domain_request_alter().
+   */
+  const DOMAIN_MATCH_NONE = 0;
+  const DOMAIN_MATCH_EXACT = 1;
+  const DOMAIN_MATCH_ALIAS= 2;
```
zerolab: needs space before =

agentrickard:
This is really interesting to me. Core doesn't seem to have this issue -- look at node.module, for instance.

Is the switch from define() to const sufficient to fix the problem?

I'm also interested in what contexts this error occurs. I have not seen it before.

kdebisschop:
I'm pretty confident that the switch from define to Const will fix the problem because the namespace-based autoloading is pretty explicit in what it does.

I do not know what triggered those errors in the log and I'm not sure how I'd go about trying to recreate them. I looked at our web access logs and found no references to /admin/config/domain in that time period. That suggests maybe our contracter ran a command via drush that was the source -- but I'm not sure what I can do to verify or narrow down further.

I did find the request ID:

> drupal-watchdog.log:Aug 12 15:01:48 staging-18244 fensitestg: |1471014108|php||||0||Notice: Undefined variable: uid in acquia_hosting_drupal_log() (line 43 of /usr/ah/lib-pub/D8-settings.functions.inc). request_id="v-b1569400-609d-11e6-a336-0eecab49a557"
> php-errors (2).log:[12-Aug-2016 15:01:48 UTC] PHP Notice:  Use of undefined constant DOMAIN_MATCH_EXACT - assumed 'DOMAIN_MATCH_EXACT' in /mnt/www/html/fensitestg/docroot/modules/domain/domain/src/DomainNegotiator.php on line 72 request_id="v-b1569400-609d-11e6-a336-0eecab49a557"

Thanks also to zerolab for finding the issues with d.o. coding standards. Obviously glad to fix those if we go forward with this pull request. Apparently, pushing a commit gets those fixes into the pull request. If that was not the right thing, let me know.

kdebisschop:
Travis CI seems to have failed because the database went away. I am looking for a way to rerun -- might just be to commit another delta

agentrickard:
Travis is being a little buggy. I'm running tests locally for verification.

My question about using `const` is actually: If we use `const` do we have to move those definitions out of the .module file?

kdebisschop:
> My question about using const is actually: If we use const do we have to move those definitions out of the .module file?

Not really - I think it would be pretty simple to do `define` statements using the  `const` values from a Constants class. 

We'd want to set things up so they were only coded in one place. I think we can go from a constants class to a define in the .module file. I'm not as confident about going from the .module to the Constants class, however.

agentrickard:
I don't think I am making myself clear. I don't want to use a Constants class. I don't see that pattern in core.

I am asking if changing the define() to const inside the existing module file will fix the problem. That's the pattern I see in core.

agentrickard:
Given that these constants are used by the Negotiator class, we should simply move them there. Code that calls the constants has to interact with the Negotiator class anyway.

kdebisschop:
I think your comment about putting the constants in the Negotiator class makes some sense. It would affect these files:
- domain/domain.api.php
- /domain/src/DomainInterface.php
- domain/src/DomainNegotiator.php
- domain/src/Entity/Domain.php
- domain_alias/domain_alias.module

Should I rework my changes to see what that approach looks like?

agentrickard:
That would be great!

domain/src/Entity/Domain.php
```diff
@@ -7,6 +7,7 @@
 use Drupal\Core\Link;
 use Drupal\Core\Url;
 use Drupal\Core\StringTranslation\StringTranslationTrait;
+use Drupal\domain\Constants;
```
agentrickard: This line can be removed.

domain/src/Tests/DomainTestBase.php
```diff
@@ -160,7 +160,7 @@ protected function drupalUserIsLoggedIn($account) {
    * @param string $field
    *   The name of the domain field used to attach to the entity.
    */
-  public function addDomainToEntity($entity_type, $entity_id, $id, $field = DOMAIN_ACCESS_FIELD) {
+  public function addDomainToEntity($entity_type, $entity_id, $id, $field = Constants::DOMAIN_ACCESS_FIELD) {
```
agentrickard: This is still Constants::

domain_source/domain_source.module
```diff
@@ -8,11 +8,6 @@
 use Drupal\Core\Entity\EntityInterface;
 
 /**
- * Defines the name of the source domain field.
- */
-define('DOMAIN_SOURCE_FIELD', 'field_domain_source');
```
agentrickard, line 13: We still need this, either as a CONST or DEFINE.

kdebisschop:
I should have done my search in codebase for relict "Constants" before pushing the commit. Sorry for the noise.

domain_source/domain_source.module
kdebisschop, line 13: As you can tell, I'm having trouble getting my fork to compare to your main line.

agentrickard:
No worries. I am grateful for the help.

Issue #259, opened 2016-08-11
Menu links do not pass entity data for Domain Source


bran:
<img width="544" alt="pathprocessormanager_php_-_lmi3dim_-____sites_lmi3dim_" src="https://cloud.githubusercontent.com/assets/111225/17605234/4a1864fa-5fce-11e6-8d23-d574809fe1eb.png">

I'm having trouble with Domain Source not rewriting menu links pointing to nodes.

For example, if a menu link points to /products which is an alias of /node/11 or entity:node/11 in the menu table, $options['entity_type'] and $options['entity'] are not set for the processor.

agentrickard:
It's probably an issue with how the data is loaded, not the proccessor order. This need to have the entity inside the $options is something I documented for Drupal core over 4 years ago, and it's frustrating to see that it still isn't consistently implemented.

I was only testing against page-based node links, so this means that we have to account for the fact that the $entity might not be present.

Personally, I think this is a core bug that needs fixing. Our handling of URLs is maddening.

agentrickard:
This is going to need a test, too.

agentrickard:
See the #262 PR for a proposed solution. After updating the code, run a cache reset.

agentrickard:
I think the work in #262 accounts for this case now.

agentrickard:
Fixed.

PR #262, opened 2016-08-16
Handle cases where the entity is not passed in via $options.

agentrickard:
See #259

### DIFF
--- a/domain_source/domain_source.module
+++ b/domain_source/domain_source.module
@@ -11,7 +11,7 @@ use Drupal\Core\Routing\TrustedRedirectResponse;
 /**
  * Defines the name of the source domain field.
  */
-define('DOMAIN_SOURCE_FIELD', 'field_domain_source');
+const DOMAIN_SOURCE_FIELD = 'field_domain_source';
 
 /**
  * Creates our fields for an entity bundle.
--- a/domain_source/domain_source.module
+++ b/domain_source/domain_source.module
@@ -116,9 +116,5 @@ function domain_source_form_submit(&$form, \Drupal\Core\Form\FormStateInterface 
   }
   $redirect_response = new TrustedRedirectResponse($url);
   $uri = Url::fromUri($url);
-  $uri->setOption('external', FALSE);
-  kint($redirect_response);
-  kint($uri);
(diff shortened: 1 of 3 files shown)


agentrickard:
Needs tests.

agentrickard:
The failure here is troubling. 

```
Pass      Browser    DomainSourceEntit  104 Drupal\domain_source\Tests\DomainSo

    Valid HTML found on "http://example.com:8080/node/add/article"

Exception User error RedirectResponseS   85 Drupal\Core\EventSubscriber\Redirec

    Redirects to external URLs are not allowed by default, use

    \Drupal\Core\Routing\TrustedRedirectResponse for it.

Pass      Browser    DomainSourceEntit  104 Drupal\domain_source\Tests\DomainSo

    POST http://example.com:8080/node/add/article returned 400 (111 bytes).

Fail      Browser    DomainSourceEntit  105 Drupal\domain_source\Tests\DomainSo

    HTTP response expected 200, actual 400
```

This suggests that the form save issued a redirect (because of the path rewrite) which was rejected by Drupal core. This redirect happens at the PathProcessorOutbound level, not the form level, so it may not be something we can account for.

I have not seen this issue when testing manually.

agentrickard:
I take that back, I do see the issue when testing manually. Drupal's form system will not allow a redirect to an 'external' URL. The recommended solution to this, TrustedRedirectResponse, appears to be incompatible with FormState::setRedirect(). 

So we either have to ignore redirects on form submit (easy), figure out how to fix this issue (hard), or issue redirects at the page-load level if someone goes to a non-canonical URL for a node. 

I am hesitant to do that last one, but it might be the best option.

zerolab:
https://www.drupal.org/node/2599342 claims you can do `$form_state->setResponse(new TrustedRedirectResponse());`

agentrickard:
FFS. What yched is saying is that you have to use setResponse(), not setRedirect(), which is a total WTF,

Thanks!

agentrickard:
Ok, this should fix up that issue nicely, then.

agentrickard:
This looks good, but still needs a test of menu links / local tabs.

agentrickard:
Should be all set now.

Issue #131, opened 2016-01-19
Allow non-ascii characters in domains


agentrickard:
This is a configuration that affects domain record validation. Some countries / systems support non-ascii characters, and should be allowed to set that as an option.

In an ideal world, all validation rules would be plugins, which could be enabled or disabled.

Issue #128, opened 2016-01-19
www prefix handling


agentrickard:
Finish the default handling of www prefix settings. (Which tell the module to treat www.example.com and example.com as the same domain.)

In older versions, we could force redirects, too, but we should drop that. If people want it, they can use Domain Alias.

Issue #46, opened 2015-03-21
Make access exempt paths configurable


agentrickard:
The list of paths in DomainAccessCheck::applies() needs to be configurable. Right now it is hardcoded to "user/*".

PR #264, opened 2016-08-28
Settings

agentrickard:
Contains configuration settings and tests for:
#128
#46
#239
#131


domain/src/Access/DomainAccessCheck.php
```diff
@@ -41,9 +41,11 @@ public function applies(Route $route) {
    * {@inheritdoc}
    */
   public function checkPath($path) {
-    $list = explode('/', $path);
-    // @TODO: This list may need to be configurable.
-    if (current($list) == 'user') {
+    $allowed_paths = \Drupal::config('domain.settings')->get('login_paths');
```
zerolab: Would be great to have the config services DI-ed. Applies for all instances in this PR.

domain/src/Form/DomainSettingsForm.php
```diff
@@ -0,0 +1,85 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  protected function getEditableConfigNames() {
+    return ['domain.settings'];
+  }
+
+  /**
+   * {@inheritdoc}
+    allow_non_ascii: false
```
zerolab: This looks like an accidental paste

domain/src/Form/DomainSettingsForm.php
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * Returns an array of settings keys.
+   */
+  public function settingsKeys() {
+    return [
+      'allow_non_ascii',
+      'www_prefix',
+      'login_paths',
+      'css_classes',
```
zerolab: `css_classes` is not present in `buildForm()`

agentrickard: I removed it since it can't be supported until token support is in.

agentrickard:
Odd, one test fail on one install -- https://travis-ci.org/agentrickard/domain/jobs/155777442#L844

domain/src/Access/DomainAccessCheck.php
agentrickard: We can add that later.

domain/src/DomainLoader.php
```diff
@@ -95,6 +95,11 @@ public function loadMultipleSorted($ids = NULL) {
    * {@inheritdoc}
    */
   public function loadByHostname($hostname) {
+    // Strip www. off the front?
+    $www = \Drupal::config('domain.settings')->get('www_prefix');
```
agentrickard: There is code replication here, but since both are public functions, I think we need to trim in both places. Best alternative would be to move this code to a helper inside the loader, since that is required by the Negotiator.

bjaxelsen:
+1

agentrickard:
Those intermittent test fails are still around. They are not related to this code.

My only real question before committing this is "Do we want the strip www. setting?" It's been around since Drupal 5, and I do consider it a core feature, but it sometimes confuses people.

PR #266, opened 2016-09-01
Fixes issue 265 -- no default settings for path exceptions.

agentrickard:


domain/src/Access/DomainAccessCheck.php
```diff
@@ -52,11 +52,11 @@ public function applies(Route $route) {
    * {@inheritdoc}
    */
   public function checkPath($path) {
-    $allowed_paths = $this->configFactory->get('domain.settings')->get('login_paths');
+    $allowed_paths = $this->configFactory->get('domain.settings')->get('login_paths', '/user/login\r\n/user/password');
```
zerolab, line 55: > The !empty should be added, because there is a case where people will disable this entirely and force login from the primary domain.

If I read the code correctly, the only way to disable this entirely is to clear the `login_paths` textarea in the settings form. As such my suggestion to provide the default `/user/login\r\n/user/password` fallback when `login_paths` is the wrong approach and only the `!empty` check is needed.

agentrickard: I disagree here for the following reasons:
- We aren't providing update hook in alpha, which would set the default value and prevent this issue, so providing a default is a workaround.
- The default value is only used if the settings key is not saved in active state.
- There is a legitimate use-case where an admin would set these paths to NULL. Doing so would disallow login on inactive domains. In that case we need the empty check.

zerolab: 👍

PR #267, opened 2016-09-01
221 views access

agentrickard:
This still needs a test, which should be pretty simple. Just a basic view in with the test module.

This was a bit tricky because there are multiple types of access checks. We can't just piggyback on the existing DomainAccessCheck, because that's a global and uses different arguments.

Here we introduce a service that only runs access checks on specific routes that require the _domain value.


agentrickard:
It's odd to me that the PR branch fails testing, but the push branch works.

agentrickard:
@zerolab this might interest you. Now with working tests.

PR #269, opened 2016-09-09
Add/Fix some type hinting

kdebisschop:
These are just so type hinting and completion works in IDE
- The hint has a typo in DomainNegotiator
- The DomainNegotiatorInterface should specify return types generally
- I added some annotations in DomainAccessManager, mostly to indicate that the EntityInterfaces are Fieldable

### DIFF
--- a/domain/src/DomainNegotiatorInterface.php
+++ b/domain/src/DomainNegotiatorInterface.php
@@ -46,6 +46,8 @@ interface DomainNegotiatorInterface {
 
   /**
    * Gets the id of the active domain.
+   *
+   * @return int|null|string
    */
   public function getActiveId();
 
@@ -55,6 +57,8 @@ interface DomainNegotiatorInterface {
    * This method is an internal method for use by the public getActiveDomain()
    * call. It is responsible for determining the active hostname of the request
    * and then passing that data to the negotiator.
+   *
+   * @return string
    */
   public function negotiateActiveHostname();
 
@@ -66,6 +70,7 @@ interface DomainNegotiatorInterface {
    *
    * @param bool $reset
    *   Reset the internal cache of the active domain.
+   * @return DomainInterface
    */
   public function getActiveDomain($reset = FALSE);
 
--- a/domain_access/src/DomainAccessManager.php
+++ b/domain_access/src/DomainAccessManager.php
@@ -41,6 +41,7 @@ class DomainAccessManager implements DomainAccessManagerInterface {
    * @inheritdoc
    */
   public function getAccessValues(EntityInterface $entity, $field_name = DOMAIN_ACCESS_FIELD) {
+    /** @var FieldableEntityInterface $entity */
     // @TODO: static cache.
     $list = array();
     // @TODO In tests, $entity is returning NULL.
@@ -66,6 +67,7 @@ class DomainAccessManager implements DomainAccessManagerInterface {
    * @inheritdoc
    */
   public function getAllValue(EntityInterface $entity) {
+    /** @var FieldableEntityInterface $entity */
     return $entity->get(DOMAIN_ACCESS_ALL_FIELD)->value;
   }
 
--- a/domain/src/DomainNegotiator.php
+++ b/domain/src/DomainNegotiator.php
@@ -95,7 +95,7 @@ class DomainNegotiator implements DomainNegotiatorInterface {
     // is required for hook_domain_request_alter().
     else {
       $values = array('hostname' => $httpHost);
-      /** @var \Drupal\domain\Entity\DomainInterface $domain */
+      /** @var \Drupal\domain\DomainInterface $domain */
       $domain = \Drupal::entityTypeManager()->getStorage('domain')->create($values);
       $domain->setMatchType(self::DOMAIN_MATCH_NONE);
     }


domain/src/DomainNegotiatorInterface.php
```diff
@@ -46,6 +46,8 @@ public function getHttpHost();
 
   /**
    * Gets the id of the active domain.
+   *
+   * @return int|null|string
```
agentrickard, line 50: This would return string or NULL.

domain_access/src/DomainAccessManager.php
```diff
@@ -41,6 +41,7 @@ public function __construct(DomainLoaderInterface $loader, DomainNegotiatorInter
    * @inheritdoc
    */
   public function getAccessValues(EntityInterface $entity, $field_name = DOMAIN_ACCESS_FIELD) {
+    /** @var FieldableEntityInterface $entity */
```
agentrickard, line 44: This variable definition doesn't match what is passed in the method.

domain_access/src/DomainAccessManager.php
```diff
@@ -66,6 +67,7 @@ public function getAccessValues(EntityInterface $entity, $field_name = DOMAIN_AC
    * @inheritdoc
    */
   public function getAllValue(EntityInterface $entity) {
+    /** @var FieldableEntityInterface $entity */
```
agentrickard, line 70: Doesn't match the variable declaration in the method.

domain_access/src/DomainAccessManager.php
```diff
@@ -90,8 +92,10 @@ public static function getDefaultValue(FieldableEntityInterface $entity, FieldDe
     switch ($entity->getEntityType()->id()) {
       case 'user':
       case 'node':
-        /** @var \Drupal\domain\DomainInterface $active */
-        if ($active = \Drupal::service('domain.negotiator')->getActiveDomain()) {
+        /** @var \Drupal\domain\DomainNegotiatorInterface $service */
+        $service = \Drupal::service('domain.negotiator');
+        $active = $service->getActiveDomain();
```
agentrickard: This two lines can be combined into one `if ($active = ...)`

domain_config/src/DomainConfigServiceProvider.php
```diff
@@ -15,7 +15,7 @@
  *
  * @see https://www.drupal.org/node/2662196#comment-10838164
  */
-class DomainConfigServiceProvider extends ServiceProviderBase implements ServiceModifierInterface {
+class DomainConfigServiceProvider extends ServiceProviderBase {
```
agentrickard: Is that implementation inherited from the base class?

agentrickard:
This is great. Thank you.

Any idea about the merge conflicts? Perhaps you need to refresh your local fork?

agentrickard:
This PR is in danger of being too long to be reviewable.

domain_access/src/Tests/DomainAccessEntityReferenceTest.php
```diff
@@ -72,6 +72,7 @@ public function testDomainAccessFieldStorage() {
     $this->assertText('Domain Access', 'Found the domain field instance.');
 
     // We expect to find 5 domain options.
+    /** @var \Drupal\domain\Entity\Domain $domain */
```
zerolab: DomainInterface

domain/domain.drush.inc
```diff
@@ -502,7 +508,8 @@ function drush_domain_default($argument) {
       $error = domain_check_response($domain, TRUE);
       if ($error) {
         drush_set_error($error);
-        drush_print(dt('You may disable this error by passing --skip_check=1.'));
+        drush_print(dt('You may dis
+    able this error by passing --skip_check=1.'));
```
zerolab: Is this an accidental line-break?

domain_access/domain_access.module
```diff
@@ -212,6 +213,10 @@ function domain_access_form_devel_generate_form_user_alter(&$form, &$form_state,
 
 /**
  * Implements hook_domain_references_alter().
+ *
+ * @param \Drupal\Core\Database\Query\SelectInterface $query
+ * @param \Drupal\user\Entity\User $account
+ * @param array $context
```
zerolab: hook implementation params should not need documenting as per https://www.drupal.org/node/1354#hookimpl

domain_access/domain_access.module
```diff
@@ -519,6 +525,8 @@ function domain_access_views_data_alter(array &$data) {
 
 /**
  * Implements hook_ENTITY_TYPE_insert().
+ *
```
zerolab: hook implementation params should not need documenting as per https://www.drupal.org/node/1354#hookimpl

zerolab:
@kdebisschop, thank you for you work on this!
As @agentrickard already mentioned, the PR is getting close to being too long for being reviewable. 

I believe the unit test is very much out of the scope for this PR and should come in a separate PR.

domain/tests/src/Unit/Access/DomainAccessCheckTest.php
```diff
@@ -0,0 +1,148 @@
+<?php
```
agentrickard: OK, now this is just overreaching within a single Pull Request.

This test addition -- though desirable -- has nothing to do with the rest of the PR and needs to be separated into it's own PR.

domain/src/DomainNegotiatorInterface.php
kdebisschop: Yeah - I saw that. The reason I did that was because DomainNegotiator::getActiveId() returns domain->id() which when I follow the train back, seems to be Entity->id() which has that signature. In truth, I don't know enough yet about the workings of the module to try to change that behavior so I thought documenting the exisiting possible return values was a sound starting point. I think the expectation is that id() returns a positive int, so changing DomainNegotiator::getActiveId() to "(int) return $this->domain->id();" and setting return type to int might be more in line with people's expectations.

domain_access/src/DomainAccessManager.php
kdebisschop: Right - but EntityInterface does not have a get() method - as far as I can tell, the only two extensions of EntityInterface that have a get() method are FieldableEntityInterface and ConfigEntityInterface - and it's pretty clear we are talking about the former here. I was not, however, able to convince myself that only FieldableEntities could ever get passed in. So I opted to try not to change the behavior.

Possibly the most correct thing to do would be to make the signature more strict. But that would risk causing errors. The safer thing to do would be to have a test on $entity to ensure that was fieldable (or at least had a get method) before calling get on it. Currently, I think this will fail somewhat badly if it gets an entity that is not fieldable.

domain_access/src/DomainAccessManager.php
kdebisschop: I believe the same response applies - currently the code will fail if ever passed an entity that is not fieldable. This type hint exposes that problem. But there may be better ways to solve the problem - but choosing among them is beyond what I can do with my understanding of the module.

domain_config/src/DomainConfigServiceProvider.php
kdebisschop: Yes, ServiceProviderBase implements the interface.

domain/tests/src/Unit/Access/DomainAccessCheckTest.php
kdebisschop: Fair enough. I need to find a way to commit without it getting merged into the outstanding pull request -- which was in fact my intent. I will reorganize. and retry.
That should take care of the merge issue. I think the best thing is to reject this pull request and try again?

kdebisschop:
Sorry for being new to really working in the git/github world. It is only now occurring to me that I should have have separate branches for the different projects I was working on. And then I guess merge the branches one at a time into 8.x-1.x -- is that the way I should approach this? Again - sorry to be just working this out now and thanks for your patience.

domain/tests/src/Unit/Access/DomainAccessCheckTest.php
zerolab: To prevent non-relevant changes from appearing in the PR, simply do not commit to the branch that the PR was opened from.

Simply branch out (i.e. create a new branch) and continue your work there.

To save this PR, use `git rebase -i` (see https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history for a nice explanation) -- basically you need to run `git rebase -i HEAD~n` (where N is the the number of commits on this branch, or the number of commits you want to change -- I counted 5)

Before that, do create a new branch off 8.x-1.x and `git cherry-pick COMMIT_HASH` for the unit test commits and the service injection commit

zerolab:
@kdebisschop it is great work, don't worry about git/GitHub. It is all fixable ;) See [some suggestions](https://github.com/agentrickard/domain/pull/269#discussion_r78301817) above.

Generally, in git branches are cheap, so create away. As on d.o. it is a good idea to contain the branch changes to a focused set.

Last but not least, for the merge conflict, you could do a couple of things:
0. (applies to all) keep your fork in sync - https://help.github.com/articles/syncing-a-fork/
1. [Rebase](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) your branch with 8.x-1.x (git rebase 8.x-1.x) and fix the merge conflict where it happens.
2. Merge 8.x-1.x in your branch and resolve the merge conflicts manually

The rebase is a cleaner method, but a bit more complicated. The merge is simpler, but results in a less clean git history.

Hope I did not confuse things more.

kdebisschop:
@zerolab - Any confusion is my fault.

I think I have successfully restored the 8.1-1.x branch to just the small set that I had originally intended to present. Now I will get to work creating branches for the unit test work and the other stuff that hemorrhaged in to the original PR.

domain_access/src/DomainAccessManager.php
agentrickard: This is a case where I don't know that commenting the variable adds any clarity, so I think we just live with the uncertainty that core hands us and don't bother annotating this variable (which is type-hinted in the line above).

domain_access/src/DomainAccessManager.php
agentrickard: We should probably change the type hint then and see if any of the tests explode.

agentrickard:
This is unfortunately outdated.

PR #270, opened 2016-09-10
125 token

agentrickard:
Adds and tests tokens,

### DIFF
--- a/domain/domain.module
+++ b/domain/domain.module
@@ -66,3 +66,17 @@ function domain_help($route_name, RouteMatchInterface $route_match) {
       return $output;
   }
 }
+
+/**
+ * Implements hook_token_info().
+ */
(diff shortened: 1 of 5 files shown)


agentrickard:
@zerolab any idea why these tests would fail on Travis? They seem to be fine locally.

agentrickard:
This is deprecated for #271, which did get a little greedy.

zerolab:
It is not directly obvious why they fail. Will have a look and hopefully have an answer in the next 1-2 days

agentrickard:
Thanks. I assume it's to do with the port 8000. I actually don't know how that gets handled by the tokens.

I suspect that Token API strips out the colon from the hostname so that example.com:8000 tokenizes to example.com8000.

Note that this code is now in #271.

PR #258, opened 2016-08-09
[WIP] Fix tests for 8.2.x, use BrowserTestBase

joshuataylor:
WIP.

This allows tests to run for 8.2.x for me.

Using to test against travis, please don't review yet.

PR #272, opened 2016-09-11
Unit test domain access check

kdebisschop:
Creates a unit test for DomainAccessCheck

Coverage is between 50 and 75% because of difficulties in providing an adequate mock for AccessResult::allowedIfHasPermissions()

### DIFF
--- a/domain/tests/src/Unit/Access/DomainAccessCheckTest.php
+++ b/domain/tests/src/Unit/Access/DomainAccessCheckTest.php
@@ -29,7 +29,22 @@ class DomainAccessCheckTest extends UnitTestCase {
       ->disableOriginalConstructor()
       ->getMock();
     $this->mockNegotiator = $negotiator;
-    $this->object = new DomainAccessCheck($negotiator);
+
+    $config = $this
+      ->getMockBuilder('Drupal\Core\Config\ImmutableConfig')
+      ->disableOriginalConstructor()->setMethods(['get'])->getMock();
+    $config->expects(static::any())->method('get')
+      ->with('login_paths')
+      ->willReturn("/user/login\n/user/password");
+
+    $configFactory = $this
+      ->getMockBuilder('Drupal\Core\Config\ConfigFactory')
+      ->disableOriginalConstructor()->setMethods(['get'])->getMock();
+    $configFactory->expects(static::any())->method('get')
+      ->willReturn($config);
+    /** @var \Drupal\Core\Config\ConfigFactory $factory */
+
+    $this->object = new DomainAccessCheck($negotiator, $configFactory);
   }
 
   /**
@@ -53,8 +68,7 @@ class DomainAccessCheckTest extends UnitTestCase {
    */
   public function testAppliesFalse($path) {
     $route = $this->getMockBuilder('\Symfony\Component\Routing\Route')
-      ->disableOriginalConstructor()
-      ->getMock();
+      ->disableOriginalConstructor()->getMock();
     $route->expects(static::once())->method('getPath')
       ->willReturn($path);
     static::assertFalse($this->object->applies($route));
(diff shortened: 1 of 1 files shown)


domain/tests/src/Unit/Access/DomainAccessCheckTest.php
```diff
@@ -0,0 +1,149 @@
...
+    $expected = AccessResult::forbidden()->setCacheMaxAge(0);
+    static::assertEquals($expected, $this->object->access($account));*/
+  }
+
+  /**
+   * @return array
+   */
+  public function providerCheckPathTrue() {
+    return [
+      ['/user/1'],
```
agentrickard, line 132: Why do these have a leading / but the ones below do not?

domain/tests/src/Unit/Access/DomainAccessCheckTest.php
```diff
@@ -0,0 +1,149 @@
...
+    static::assertFalse($this->object->checkPath($path));
+  }
+
+  /**
+   * @todo Find a way to mock AccessResult::allowedIfHasPermissions to complete test
+   *
+   * If AccessResult::allowedIfHasPermissions cannot be done by mock, then
+   * it will need to be done as a web test case. Maybe an example in comment\Unit\Entity\CommentLockTest
+   */
+  public function testAccess() {
```
agentrickard, line 109: This is going to be interesting to mock, since the DomainAccessCheck hinges on the status of the domain record and the permissions of the user.

I would expect two mock domain objects to test against -- one where status = 1 and one where status = 0.

I don't know that mocking the Negotiator gets us what we want. It might be easier to mock the DomainLoader?

agentrickard:
My concern here is that most of the features require browser testing, since we're dealing with http requests. 

I'd rather spend energy on moving to the new BrowserTestBase than chasing unit tests.

domain/tests/src/Unit/Access/DomainAccessCheckTest.php
kdebisschop: I wanted to make sure that the test cases captured places along the edge of expected behavior. So, if someone got and started pulling path from a different source, it would turn up. Moot issue now because the changes to the DomainAccessCheck that switched to to a list from config makes those tests entirely incorrect. I have updated the tests accordingly.

domain/tests/src/Unit/Access/DomainAccessCheckTest.php
kdebisschop: Going along with your comment below, I tend to agree that there is a point of vanishing returns. I may look at whether the the non-trivial tests can be done as a KernelTest -- might be a good deal more portable. (But I will have to become familiar with kernel tests to do that). I have deleted all that commented code and currently don't plan to try implementing those user checks as a unit test. Depending on what I find about kernel tests, the tests here might just move to a kernel test.

agentrickard:
I was on a plane yesterday, but see also #258 for the move to the new test framework.

kdebisschop:
Looked at #258 and the link into https://www.drupal.org/node/2735005. In that article I see:

> In the meantime open child issues to convert web tests for each core module. While doing so we should also check if the test cases actually require a browser or can be converted to unit/kernel tests.

My understanding is that all of BrowserTest, KernelTest, and UnitTest were to be supprted going forward, but that the lowest overhead should be used when possible. Which would be unit test for the test cases that do not need user info, etc, kernel test for things that need something like a drupal runtime, and browser only for things that actually interact with the browser.

All are PHPunit, so should be driven by travis. Just with less overhead if you don't have the browser.

Am I misunderstanding that?

Am I not seeing what you are driving at?

agentrickard:
I'm saying that since much of the  behavior requires HTTP request/response, Unit testing may not be practical.

I also don't have time to get up to speed on Unit and Kernel tests, so the conversion from WebTestBase to BrowserTestBase is more important to me personally.

agentrickard:
Given the changes to the travis test runner made in https://github.com/agentrickard/domain/pull/286/commits/2412c52f8c729dee90e15a21be29603381d690d5, we likely need to add lines to the test runner configuration to explicitly call these unit tests.

PR #273, opened 2016-09-11
Unit test domain validator

kdebisschop:
Some work by agentrickard duplicated what I was doing to use service injection and create a unit test. I've tried to keep that as much as possible, reordering my parameters to match, etc. One change in 8.1-1.x that I think was incorrect was to switch from Guzzle\Client to ClientInterface -- I think get() is a magic method provided by the actual Guzzle Client and not the interface. So if that type is change to the interface, there is no assurance we will still have get().

I also choose to get the domain config once in the constructor instead of storing the whole configuration - it seemed to produce cleaner code, but could be changed later without affecting the interface.

There was a note about splitting the validate function (though the rationale was not given) It made sense to me that half was stuff related to the domain config and the other half a set of string inspections that pretty much any hostname validator would do. So I assumed that rationale and made the split, which sets up the code to use a generic validator instead of writing one custom.

Lastly, I added a optional string path parameter to checkResponse. Certainly part of the reason was that drupal_get_path is not playing well with unit test operation. But it also made sense that one could want to specify a path other than test.png for many reasons. So it felt OK, but it too was a judgement call.

### DIFF
--- a/domain/src/DomainValidator.php
+++ b/domain/src/DomainValidator.php
@@ -3,10 +3,12 @@
 namespace Drupal\domain;
 
 use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Extension\ModuleHandlerInterface;
 use Drupal\Core\StringTranslation\StringTranslationTrait;
 use Drupal\Component\Utility\Unicode;
-use GuzzleHttp\ClientInterface;
+use Drupal\Core\StringTranslation\TranslationManager;
+use GuzzleHttp\Client;
 use GuzzleHttp\Exception\RequestException;
 
 /**
@@ -15,6 +17,16 @@ use GuzzleHttp\Exception\RequestException;
 class DomainValidator implements DomainValidatorInterface {
 
   use StringTranslationTrait;
+
+  /**
+   * @var \Drupal\Core\Config\ImmutableConfig
+   */
+  protected $config;
+
+  /**
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
 
   /**
    * The module handler.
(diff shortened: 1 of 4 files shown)


domain/src/DomainValidator.php
```diff
@@ -3,10 +3,11 @@
 namespace Drupal\domain;
 
 use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Extension\ModuleHandlerInterface;
 use Drupal\Core\StringTranslation\StringTranslationTrait;
 use Drupal\Component\Utility\Unicode;
-use GuzzleHttp\ClientInterface;
```
agentrickard, line 9: My understanding is that it is generally best to call the Interface here, which guards against classes being swapped out.

If get() is not part of the Interface, then this change is correct and we have a hard dependency.

The rest of this looks great.

agentrickard:
I had been thinking of splitting the validator into methods or plugins for each of the specific tests. I have no real interest in abstracting to a library that I have to maintain outside of Drupal.

The real question is whether we should allow some of these rules to be disabled. If not, the current level of abstraction in this PR is fine.

domain/src/DomainValidator.php
zerolab: Core uses `GuzzleHttp\ClientInterface`. e.g. http://cgit.drupalcode.org/drupal/tree/core/modules/update/src/UpdateFetcher.php

kdebisschop: Interesting. I checked the Guzzle code again, and get() is a magic method that is only in the client and not in the interface. I cannot find a way to believe core isn't wrong.

One way to follow the (very desirable) practice of coding to an interface and still have code that is consistent with the interface is to forgo get() and use request('get'). I have changed the code to use request() instead of get(). Do you think that is a reasonable approach?

agentrickard: Yes. Magic methods are bad for business.

kdebisschop:
On how to split the validator, you say "the real question is whether we should allow some of these rules to be disabled" When I parse that down, it sounds like we might have a set of check that always run, and then a method for each check that can be disabled (or modified) by a Drupal configuration setting. Is that what you're thinking? 

If so, would that mean the current check for allow_non_ascii and www_prefix should be their own methods that can add an element to the error array?

Since we have settled on Guzzle request() so we can code to interface, I have remove the last bits that referred to get(). Other that the question about how to split, is there anything else I should address in this PR?

agentrickard:
That is a correct assessment of the "separating validation" but that work doesn't need to go in this PR.

If this PR just handles the current code, fixes Giuzzle, and has a working test, that's fine.

We can move the abstraction/configuration of validators to another task -- and those would also apply to Aliases. That abstraction is not required for beta release.

domain/src/DomainValidator.php
```diff
@@ -26,103 +37,72 @@ class DomainValidator implements DomainValidatorInterface {
...
+    if (! $this->config->get('allow_non_ascii') && ! preg_match('/^[a-z0-9:.-]*$/i', $hostname)) {
+      $error_list[] = $this->t('Only alphanumeric characters, dashes, and a colon are allowed.');
     }
+
     // Check for 'www' prefix if redirection / handling is
     // enabled under global domain settings.
     // Note that www prefix handling must be set explicitly in the UI.
     // See http://drupal.org/node/1529316 and http://drupal.org/node/1783042
-    if ($this->configFactory->get('domain.settings')->get('www_prefix') && (substr($hostname, 0, strpos($hostname, '.')) == 'www')) {
+    if ($this->config->get('www_prefix') && 0 === strpos($hostname, '.www')) {
```
agentrickard: The strpos statement here is backward. The check condition is `www.` not `.www`. I also don't use the "value" first if ordering.

domain/src/DomainValidator.php
```diff
@@ -26,103 +37,72 @@ class DomainValidator implements DomainValidatorInterface {
...
-    if (!$non_ascii) {
-      $pattern = '/^[a-z0-9\.\-:]*$/i';
-      if (!preg_match($pattern, $hostname)) {
-        $error_list[] = $this->t('Only alphanumeric characters, dashes, and a colon are allowed.');
-      }
-    }
-    // Check for lower case.
-    if ($hostname != Unicode::strtolower($hostname)) {
-      $error_list[] = $this->t('Only lower-case characters are allowed.');
+    if (! $this->config->get('allow_non_ascii') && ! preg_match('/^[a-z0-9:.-]*$/i', $hostname)) {
```
agentrickard: No spaces between the ! negation and the variable.

domain/src/DomainValidator.php
```diff
@@ -26,103 +37,72 @@ class DomainValidator implements DomainValidatorInterface {
...
     }
 
     // Check existing domains.
-    $domains = \Drupal::entityTypeManager()
-      ->getStorage('domain')
-      ->loadByProperties(array('hostname' => $hostname));
-    foreach ($domains as $match) {
-      if ($match->id() != $domain->id()) {
-        $error_list[] = $this->t('The hostname is already registered.');
+    try {
```
agentrickard, line 92: Smart.

domain/src/DomainValidator.php
```diff
@@ -26,103 +37,72 @@ class DomainValidator implements DomainValidatorInterface {
...
+      $error_list[] = $this->t('Could not open domain storage');
+    }
+
     // Allow modules to alter this behavior.
-    \Drupal::moduleHandler()->invokeAll('domain_validate', array($error_list, $hostname));
+    $this->moduleHandler->invokeAll('domain_validate', array($error_list, $hostname));
 
     // Return the errors, if any.
-    if (!empty($error_list)) {
+    if (count($error_list)) {
```
agentrickard: I prefer !empty here for clarity.

domain/src/DomainValidatorInterface.php
```diff
@@ -28,11 +28,13 @@ public function validate(DomainInterface $domain);
    *
    * @param \Drupal\domain\DomainInterface $domain
    *   A domain record.
+   * @param string $test_path
+   *   A path on the domain to test.
```
agentrickard: I would add `If not set, the request will test a file in a known location.`

domain/tests/src/Unit/DomainValidatorTest.php
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * @group domain
+ */
+class DomainValidatorTest extends UnitTestCase {
+
+  /** @var \Drupal\domain\DomainValidator */
+  private $object;
+
+  /** @var \GuzzleHttp\Client */
+  public $client;
```
agentrickard: For my own curiosity, why is $client public here?

domain/tests/src/Unit/DomainValidatorTest.php
```diff
@@ -0,0 +1,126 @@
...
+    $mock->expects(static::any())->method('getHostname')
+      ->willReturn($host);
+    $errors = $this->object->validate($domain);
+    static::assertNotEquals((bool) $result, (bool) count($errors));
+  }
+
+  /**
+   * @return array
+   */
+  public function provider() {
```
agentrickard: Is `provider()` a standard naming convention for PHPUnit? If not, this seems like an odd name. I would think that `expectedResult()` is more clear. So it's not entirely clear how this is invoked.

agentrickard (changes requested):
Final revisions / clarifications

domain/tests/src/Unit/DomainValidatorTest.php
kdebisschop: The convention is providerAspectUnderTest provides data for testAspectUnderTest. In starting out, I had tested the static methods directly before making making it private. When I ran the same test data through both the public method and the (now private) internal method, I left it more generic. I will update to match the test name as is usually the case.

expectedResult() seems less clear because the data provider contains both the result and the data that produce the result. I've never seen it used as a name for a data provider.

The name is only a convention, The @dataProvider annotation invokes the method, so any method name works as long as it is the name in the annotation.

domain/src/DomainValidatorInterface.php
kdebisschop: Makes sense, but maybe can be tighter wording to stay on one line. Also needs the (optional) annotation.

domain/src/DomainValidator.php
kdebisschop: typo on '.www' is a bit annoying -- that was one of the specific cases I meant to cover in the unit test to make sure I got it right. But got block by problems injecting string translation. Good news is that I got the translation service injection to work. Which means nearly 100% coverage of the class in the unit test. But it does make the PR a bit bigger.

domain/tests/src/Unit/DomainValidatorTest.php
agentrickard: Thanks for that, the annotation trick is one that would have given me fits.

agentrickard:
I'm about to head out for vacation and will be offline until next Tuesday, when I resurface at DrupalCON Dublin.

domain/src/DomainValidator.php
kdebisschop: Will fix. Thanks.

domain/src/DomainValidator.php
kdebisschop: Will fix,

domain/tests/src/Unit/DomainValidatorTest.php
kdebisschop: Not required to be public. Just unsure of myself as I started working with dataProvider (which does need to be public)

PR #274, opened 2016-09-11
Resolve places where subclasses mask parts of parent class

kdebisschop:
Most of these seem to be innocent in that the subclass does not actually use the property it masked. So the accessors on the parent class would access the original properties and the behavior so far was probably no different than what you would expect. So they are about protecting from future undocumented surprises.

It's worth calling out DomainControllerBase, however. In that class, the EntityTypeManager was being written into a local version of the EntityManager. Both are properties in the parent class. I looked at the code and I think that because DomainAliasController uses \Drupal::entityTypeManager(), that error I think had no consequences either. But I think we would have started seeing errors if we had been using dependency injection for the EntityTypeManager service. It's worth someone else looking at the code to see if you agree with my analysis.

### DIFF
--- a/domain/src/DomainListBuilder.php
+++ b/domain/src/DomainListBuilder.php
@@ -3,8 +3,6 @@
 namespace Drupal\domain;
 
 use Drupal\Core\Config\Entity\DraggableListBuilder;
-use Drupal\Core\Entity\EntityTypeInterface;
-use Drupal\Core\Entity\EntityStorageInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Url;
@@ -15,14 +13,9 @@ use Drupal\Core\Url;
 class DomainListBuilder extends DraggableListBuilder {
 
   /**
-   * Provide class-specific value of $entitiesKey.
-   *
    * {@inheritdoc}
    */
-  public function __construct(EntityTypeInterface $entity_type, EntityStorageInterface $storage) {
-    $this->entitiesKey = 'domains';
-    parent::__construct($entity_type, $storage);
-  }
+  protected $entitiesKey = 'domains';
 
   /**
    * {@inheritdoc}
--- a/domain/src/DomainListBuilder.php
+++ b/domain/src/DomainListBuilder.php
@@ -13,9 +13,14 @@ use Drupal\Core\Url;
 class DomainListBuilder extends DraggableListBuilder {
 
   /**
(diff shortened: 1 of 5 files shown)


domain/src/DomainListBuilder.php
```diff
@@ -13,9 +13,14 @@
 class DomainListBuilder extends DraggableListBuilder {
 
   /**
+   * Provide class-specific value of $entitiesKey.
+   *
    * {@inheritdoc}
    */
-  protected $entitiesKey = 'domains';
```
agentrickard, line 18: Don't we still need to declare this property?

agentrickard:
Except for that one question, the rest of this looks good. The $entity_manger / $entity_type_manager issue is due to a change in core during the dev cycle. 

$entity_type_manager is more correct.

domain/src/DomainListBuilder.php
kdebisschop: I think I can show that we do not need to declare the property with the code below:

`<?php

abstract class MyParent {
  protected $maskedData = 'masked';

  protected $parentData = 'parent';

  public function __construct() {
  }

  protected function test() {
    echo 'In ', **CLASS**, ": $this->maskedData\n";
    echo 'In ', **CLASS**, ": $this->parentData\n";
  }

  public function reset() {
    $this->maskedData = $this->parentData = 'reset';
  }
}

class MyChild extends MyParent {
  protected $maskedData = 'changed by child definition';

  public function __construct() {
    parent::__construct();
    $this->parentData = 'changed by constructor';
  }

  public function test() {
    parent::test();
    echo 'In ', **CLASS**, ":  $this->maskedData\n";
    echo 'In ', **CLASS**, ":  $this->parentData\n";
  }
}

$child = new MyChild();
$child->test();
$child->reset();
$child->test();
`

I have one had a co-worker get bit by the bug this sort of thing can cause, so I was hoping to be able to demonstrate that as well. But have so far failed. Conceptually, what we want is for there to be one variable, which is part of the motivation - for clarity of code. On the other hand, there are several places in core that do use the original construction. Given that fact, I spent the last hour trying to decide myself if it actually make sense to make this change. It works, I think it seems safe, but being different than core with a solid reason does not seem like a virtue.

I'm going to look at your other comments and come back to this.

kdebisschop:
I have reverted the constructor instead of protected $entitiesKey = 'domains'; since there are similar examples in other modules and I could not demonstrate an actual impact unless the property was made private. Going with the thought that "being different than core without a solid reason does not seem like a virtue" I believe that resolves the pending questions on this PR.

PR #275, opened 2016-09-12
Url class needs namespace here

kdebisschop:
Url class needs namespace here because this file is in root namespace and has no use statements

@return annotation indicate there is a return in PHPDoc. This cause problems with IDEs an other systems that infer interface from those annotations. So remove annotations but leave comment.


agentrickard:
This test fails are not relevant to the PR.

That @return syntax appears frequently in my code, and likely needs a lot of cleanup.

Issue #277, opened 2016-10-19
UI to define a homepage for each domain


dmouse:
I want to customize the homepage in each domain, my first approach it defines the value on each domain in `Drupal\domain\DomainForm` to create the homepage field (image below)

![image](https://cloud.githubusercontent.com/assets/198571/19540191/5c9befc6-9623-11e6-9e14-5e983aae18e8.png)

Internally We need to create a new `PathProcessor` to read information for each domain

``` php
<?php

namespace Drupal\domain\PathProcessor;

/**
 * Processes the inbound path by resolving it to the front page if empty. 
 */
class PathProcessorDomainHomepage implements InboundPathProcessorInterface, OutboundPathProcessorInterface { 

}
```

What do you think about it?

dmouse:
My first approach is using nodes, maybe this can be changed for any path into a Drupal site.
See this pull request #278

dmouse:
Do you have any feedback for me?

agentrickard:
Yes, I put it in the PR just now.

PR #278, opened 2016-10-20
UI to define homepage #277

dmouse:
follow the conversation on #277

### DIFF
--- a/domain/domain.services.yml
+++ b/domain/domain.services.yml
@@ -44,10 +44,9 @@ services:
     tags:
       - { name: persist }
     arguments: ['@module_handler', '@config.factory', '@http_client']
-
   domain.path_processor_homepage:
     class: Drupal\domain\PathProcessor\PathProcessorDomainHomepage
     tags:
       - { name: path_processor_inbound, priority: 200 }
       - { name: path_processor_outbound, priority: 200 }
-    arguments: ['@config.factory']
+    arguments: ['@config.factory', '@domain.loader']
--- a/domain/domain.services.yml
+++ b/domain/domain.services.yml
@@ -44,3 +44,10 @@ services:
     tags:
       - { name: persist }
     arguments: ['@module_handler', '@config.factory', '@http_client']
+
+  domain.path_processor_homepage:
+    class: Drupal\domain\PathProcessor\PathProcessorDomainHomepage
+    tags:
+      - { name: path_processor_inbound, priority: 200 }
+      - { name: path_processor_outbound, priority: 200 }
+    arguments: ['@config.factory']
--- a/domain/src/DomainForm.php
+++ b/domain/src/DomainForm.php
@@ -158,13 +158,24 @@ class DomainForm extends EntityForm implements ContainerInjectionInterface {
   /**
    * {@inheritdoc}
    */
-  public function validate(array $form, FormStateInterface $form_state) {
+  public function validateForm(array &$form, FormStateInterface $form_state) {
     $entity = $this->buildEntity($form, $form_state);
     $errors = $this->domainValidator->validate($entity);
     if (!empty($errors)) {
       $form_state->setErrorByName('hostname', $errors);
     }
+
+    if (!$this->pathValidator->isValid($form_state->getValue('homepage'))) {
+      $form_state->setErrorByName('homepage',
+        $this->t("The path '%path' is either invalid or you do not have access to it.", array(
+          '%path' => $form_state->getValue('homepage')
+          )
+        )
+      );
+    }
   }
+
+
 
   /**
    * {@inheritdoc}
@@ -180,8 +191,6 @@ class DomainForm extends EntityForm implements ContainerInjectionInterface {
     $domain->save();
     $form_state->setRedirect('domain.admin');
   }
-
-
 
   /**
    * {@inheritdoc}
--- a/domain/src/DomainForm.php
+++ b/domain/src/DomainForm.php
@@ -137,11 +137,11 @@ class DomainForm extends EntityForm implements ContainerInjectionInterface {
       '#description' => $this->t('If a URL request fails to match a domain record, the settings for this domain will be used. Only one domain can be default.'),
     );
     $form['homepage'] = [
-      '#type' => 'entity_autocomplete',
+      '#type' => 'textfield',
       '#title' => $this->t('homepage'),
       '#target_type' => 'node',
       '#tags' => FALSE,
-      '#default_value' => Node::load($domain->getHomepage()),
+      '#default_value' => $domain->getHomepage(),
       '#description' => $this->t('Define the homepage'),
     ];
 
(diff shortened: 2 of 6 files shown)


domain/domain.services.yml
```diff
@@ -44,3 +44,10 @@ services:
     tags:
       - { name: persist }
     arguments: ['@module_handler', '@config.factory', '@http_client']
+
+  domain.path_processor_homepage:
+    class: Drupal\domain\PathProcessor\PathProcessorDomainHomepage
+    tags:
+      - { name: path_processor_inbound, priority: 200 }
+      - { name: path_processor_outbound, priority: 200 }
+    arguments: ['@config.factory']
```
gnuget: Please add a new line here.

domain/src/DomainInterface.php
```diff
@@ -199,4 +199,14 @@ public function getMatchType();
    */
   public function createDomainId();
 
+  /**
```
gnuget, line 202: Can you please add documentation to these methods? Thanks!

domain/src/Entity/Domain.php
```diff
@@ -474,4 +482,12 @@ public function getMatchType() {
     return $this->matchType;
   }
 
+  public function getHomepage() {
```
gnuget, line 485: here we also need to include the docblocks.  adding {@inheritdoc} would be enough (if you document the interface)

domain/src/PathProcessor/PathProcessorDomainHomepage.php
```diff
@@ -0,0 +1,79 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function processInbound($path, Request $request) {
+
+    if ($path === '/') {
+      $hostname = $request->getHost();
+
+      /** @var \Drupal\domain\DomainInterface $domain */
```
gnuget: all inline comments must use  //   not /*\* */

domain/src/PathProcessor/PathProcessorDomainHomepage.php
```diff
@@ -0,0 +1,79 @@
...
+      $domain = \Drupal::service('domain.loader')->loadByHostname($hostname);
```
gnuget: not sure but I think If this is a service it would be a better idea to inject the dependency instead to use  the \Drupal::service  static method.

domain/src/DomainForm.php
```diff
@@ -4,11 +4,65 @@
...
+
+  /**
+   * DomainForm constructor.
+   *
+   * @param \Drupal\domain\DomainValidator $domainValidator
+   * @param \Drupal\domain\DomainLoader $domainLoader
+   * @param \Drupal\domain\DomainCreator $domainCreator
+   * @param PathValidatorInterface $pathValidator
+   */
+  public function __construct(DomainValidator $domainValidator, DomainLoader $domainLoader, DomainCreator $domainCreator, PathValidatorInterface $pathValidator) {
```
gnuget, line 48: all the variables in the method signature must be snake case.

Check this comment :-) https://www.drupal.org/node/2303963#comment-10111290

domain/domain.services.yml
```diff
@@ -44,3 +44,9 @@ services:
     tags:
       - { name: persist }
     arguments: ['@module_handler', '@config.factory', '@http_client']
+  domain.path_processor_homepage:
```
agentrickard, line 47: Does this replace the existing path processor in Domain Config?

domain/src/DomainForm.php
```diff
@@ -4,11 +4,65 @@
...
+use Drupal\domain\DomainLoader;
+use Drupal\domain\DomainValidator;
+use Drupal\domain\DomainCreator;
+use Drupal\Core\Path\PathValidatorInterface;
 
 /**
  * Base form for domain edit forms.
  */
-class DomainForm extends EntityForm {
+class DomainForm extends EntityForm implements ContainerInjectionInterface {
```
agentrickard, line 18: I didn't know how to do this. If nothing else, this part of the patch should get in.

domain/src/DomainForm.php
```diff
@@ -80,27 +136,47 @@ public function form(array $form, FormStateInterface $form_state) {
       '#default_value' => $domain->isDefault(),
       '#description' => $this->t('If a URL request fails to match a domain record, the settings for this domain will be used. Only one domain can be default.'),
     );
-    $required = \Drupal::service('domain.validator')->getRequiredFields();
+    $form['homepage'] = [
```
agentrickard, line 139: Two things I don't like here:
- Restricting the page to only nodes. (The last site I built has a View for a homepage.)
- This contradicts the existing logic that we have in Domain Config, which will cause confusion, because that will always present an option for overriding the homepage path (and has it's own Path Processor as well.)

My problem here is that this is a one-off approach to the problem of a UI, and I'd like to avoid that if possible.  We're also saving this as a config setting on the domain, not overriding the actual homepage setting, which I think will lead to confusion.

I can see a stronger case for perhaps adding a Domain Config UI that adds these form elements, but I would prefer that they be tied to the domain.config overrides, not the domain config entity itself. (Essentially we're mixing two systems here, which we should try to avoid.)

agentrickard (commented):
I would prefer to see an approach that leverages the config system, not the domain entity itself. 

We should be able to do this in some way that doesn't require altering the base domain entity and mixing the concepts of config entity and config object.

domain/domain.services.yml
dmouse: No, in the Domain Config exist only a `RouteProvider` and this is a `Drupal\Core\PathProcessor`, the path [processor is used](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21PathProcessor%21PathProcessorManager.php/function/PathProcessorManager%3A%3AprocessInbound/8.2.x) to [check the frontpage, aliases, etc.](https://api.drupal.org/api/drupal/namespace/Drupal%21Core%21PathProcessor/8.2.x).
This service is basically a replacement to `PathProcessorFront`

domain/src/DomainForm.php
dmouse: :smile:

domain/src/DomainForm.php
dmouse: > Restricting the page to only nodes. (The last site I built has a View for a homepage.)

Done, I changed to use paths instead Nodes

An option would be when You select a domain like default you don't change the homepage in the domain entity,
https://github.com/agentrickard/domain/pull/278/files#diff-f75f10f3e7bbd9a5dc37509fa77fc66dR54

domain/src/DomainForm.php
dmouse: :cry: Ok

agentrickard:
I don't think you should continue down this path. 

Loading this data onto the domain entity is a violation of the module design. This kind of thing should be handled by Domain Config and not special-cased just for one variable.

I understand that we might consider the homepage a special case, but I don't think that's a strong enough reason to violate the separation of concerns.

If you want to do something like this in a separate module, that's fine, but this approach is not going in to the main module.

PR #283, opened 2016-11-04
282 form

agentrickard:
Solves the problem described by https://www.drupal.org/node/2815951

@TODO 

* Needs a test
* Move new functions to the manager service.


agentrickard:
This does not appear to be testable with WebTestCase. May need the new browser tests.

domain/src/DomainFieldManager.php
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Generic base class for handling hidden field options.
+ *
+ * Since domain options are restricted for various forms (users, nodes, source)
+ * we have a base class for handling common use cases. The details of each
+ * implementation are generally handled by a subclass and invoked within a
+ * hook_form_alter().
+ */
+class DomainFieldManager {
```
agentrickard: This can probably be the base class, and other classes can extend it. That would remove the need to call the $service dynamically, which is fugly code.

agentrickard: So we can make a DomainFieldManagerInterface, implement it here. And then abstract DomainAccessFieldManager and DomainSourceFieldManager into services that extend the base class implementation.

agentrickard:
This is looking good. TODO 
- Merge recent changes from 8.x-1.x
- Write a BrowserTest for the form based on the notes at https://www.drupal.org/docs/8/phpunit/phpunit-browser-test-tutorial

PR #285, opened 2016-11-07
Domain config overrides in settings.php

zerolab:
Based on #284.

This PR enables `DomainConfigOverrider` to pick up configuration changes in settings.php

Somehow it feels more natural to have `$config['domain.config'][$domain_id][$langcode][$config_name];` or `$config['domain.config'][$domain_id][$config_name];`, but for now, `$config["domain.config.{$domain_id}.{$langcode}.{$config_name}"];` will have to do

ToDo:
- [x] tests

### DIFF
--- a/domain/src/Tests/DomainIResponseTest.php
+++ b/domain/src/Tests/DomainIResponseTest.php
@@ -1,14 +1,20 @@
 <?php
 
 namespace Drupal\domain\Tests;
+
 use Drupal\Component\Render\FormattableMarkup;
 
 /**
  * Tests domain record HTTP response.
  *
+ * Note: Class is called DomainIResponseTest on purpose as for some
+ * unexplained reason when it is DomainResponseTest and runs after
+ * DomainNegotiatorTest, it plain fails due to timeout (no matter how high it
+ * gets set in DomainValidator::checkResponse().
+ *
  * @group domain
  */
-class DomainResponseTest extends DomainTestBase {
+class DomainIResponseTest extends DomainTestBase {
 
   /**
    * Tests that a domain response is proper.
--- a/domain_config/src/DomainConfigOverrider.php
+++ b/domain_config/src/DomainConfigOverrider.php
@@ -94,6 +94,14 @@ class DomainConfigOverrider implements ConfigFactoryOverrideInterface {
         elseif ($override = $this->storage->read($config_name['domain'])) {
           $overrides[$name] = $override;
         }
+
+        // Apply any settings.php overrides.
+        if (isset($GLOBALS['config'][$config_name['langcode']])) {
+          $overrides[$name] = $GLOBALS['config'][$config_name['langcode']];
+        }
+        elseif (isset($GLOBALS['config'][$config_name['domain']])) {
+          $overrides[$name] = $GLOBALS['config'][$config_name['domain']];
+        }
       }
     }
     return $overrides;
--- a/domain_config/src/Tests/DomainConfigOverriderTest.php
+++ b/domain_config/src/Tests/DomainConfigOverriderTest.php
@@ -54,6 +54,34 @@ class DomainConfigOverriderTest extends DomainConfigTestBase {
   }
 
   /**
+   * Tests that domain-specific variable overrides in settings.php works.
+   */
+  public function testDomainConfigOverriderFromSettings() {
+    // Set up overrides.
+    $settings = [];
+    $settings['config']['domain.config.one_example_com.en.system.site']['name'] = (object) [
+      'value' => 'First',
+      'required' => TRUE,
+    ];
+    $settings['config']['domain.config.four_example_com.system.site']['name'] = (object) [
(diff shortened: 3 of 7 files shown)


agentrickard:
Is this at all testable?

zerolab:
I think so. Should be a matter of setting the correct `$GLOBALS['config']...` overrides in  in `setUp()`.

Hope to provide it tonight, if all goes well

agentrickard:
Right, since they are just globals, the test can be added to the existing DomainConfigOverriderTest case.

zerolab:
I have to admit the test for this beat me into submission :(

Cannot inject them through `$_GLOBALS` no matter what I do. Tried anything from `$this->refreshVariables();`, to rebuilding the container and things in between.

agentrickard:
There don't appear to be any core tests for this behavior.

domain_config/src/DomainConfigOverrider.php
```diff
@@ -94,6 +94,14 @@ public function loadOverrides($names) {
         elseif ($override = $this->storage->read($config_name['domain'])) {
           $overrides[$name] = $override;
         }
+
+        // Apply any settings.php overrides.
+        if (isset($GLOBALS['config'][$config_name['langcode']])) {
```
agentrickard, line 99: Are we certain that $config_name['language'] is set in all cases?

zerolab:
There are `KernelTest`s that do it: [ConfigOverrideTest](http://cgit.drupalcode.org/drupal/tree/core/tests/Drupal/KernelTests/Core/Config/ConfigOverrideTest.php).

I found the solution in [ConfigFormOverrideTest](http://cgit.drupalcode.org/drupal/tree/core/modules/config/src/Tests/ConfigFormOverrideTest.php) which uses `$this->writeSettings` to override some settings.

Pushed. They are all green locally. Let's see if Travis agrees (may need to look at it separately)

domain_config/src/DomainConfigOverrider.php
zerolab: `getDomainConfigName()` always returns `'langcode' => 'domain.config.' . $domain->id() . '.' . $this->language->getId() . '.' . $name,`, so $config_name['langcode'] will have a value

domain/src/Tests/DomainIResponseTest.php
```diff
@@ -1,14 +1,20 @@
 <?php
 
 namespace Drupal\domain\Tests;
+
 use Drupal\Component\Render\FormattableMarkup;
 
 /**
  * Tests domain record HTTP response.
  *
+ * Note: Class is called DomainIResponseTest on purpose as for some
```
agentrickard: typo

domain_config/src/Tests/DomainConfigOverriderTest.php
```diff
@@ -54,6 +54,34 @@ public function testDomainConfigOverrider() {
...
+    $settings = [];
+    $settings['config']['domain.config.one_example_com.en.system.site']['name'] = (object) [
+      'value' => 'First',
+      'required' => TRUE,
+    ];
+    $settings['config']['domain.config.four_example_com.system.site']['name'] = (object) [
+      'value' => 'Four overridden in settings',
+      'required' => TRUE,
+    ];
+    $this->writeSettings($settings);
```
agentrickard, line 70: Does this actually test the settings.php behavior? Doesn't this write mean that we're loading from config storage?

zerolab: As per function [docbloc](http://cgit.drupalcode.org/drupal/tree/core/modules/simpletest/src/WebTestBase.php?h=8.1.x#n920)

``` php
  /**
   * Rewrites the settings.php file of the test site.
   *
   * @param array $settings
   *   An array of settings to write out, in the format expected by
   *   drupal_rewrite_settings().
   *
   * @see drupal_rewrite_settings()
   */
  protected function writeSettings(array $settings) {
    include_once DRUPAL_ROOT . '/core/includes/install.inc';
    $filename = $this->siteDirectory . '/settings.php';
    // system_requirements() removes write permissions from settings.php
    // whenever it is invoked.
    // Not using File API; a potential error must trigger a PHP warning.
    chmod($filename, 0666);
    drupal_rewrite_settings($settings, $filename);
  }
```

and `drupal_write_settings()` does the actual writing to file, invalidates the opcode cache and loads the new settings in a `Settings` object.

So I believe we are testing the correct thing.

agentrickard: I see. Thank you. I was reading that wrong.

domain/src/Tests/DomainIResponseTest.php
agentrickard: Lets rename this DomainGetResponse and see if that still works.

zerolab: Done. All tests are green locally

zerolab:
Thank you!

PR #286, opened 2016-11-11
134 edit

agentrickard:
Checking status of this work.


agentrickard:
These tests fail because the Entity Reference System does not give enough information to know _which_ field is being altered by a given request.

That leads to a situation where hook_domain_references_alter() doesn't know what field it is responding to, so the conditions for both sets are applied and the query returns empty.

So I think that means we need a new EntityReferenceSelection plugin for this field, because there is no obvious way to coerce the proper context.

domain/src/DomainAccessControlHandler.php
```diff
@@ -18,25 +18,61 @@ class DomainAccessControlHandler extends EntityAccessControlHandler {
...
+   *
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   The entity to retrieve field data from.
+   * @param \Drupal\Core\Session\AccountInterface
+   *   The user account.
+   *
+   * @return boolean
+   */
+  public function isDomainAdmin(EntityInterface $entity, AccountInterface $account) {
+    $user = \Drupal::entityTypeManager()->getStorage('user')->load($account->id());
```
zerolab: candidate for DI

domain/src/DomainAccessControlHandler.php
```diff
@@ -18,25 +18,61 @@ class DomainAccessControlHandler extends EntityAccessControlHandler {
...
+   * @param string $field_name
+   *   The name of the field that holds our data.
+   *
+   * @return array
+   *   An array keyed by domain entity id.
+   *
+   * @TODO: Refactor this code so that it can be shared by other modules.
+   */
+  public function getAdminValues(EntityInterface $entity, $field_name = DOMAIN_ADMIN_FIELD) {
+    return \Drupal::service('domain.element_manager')->getFieldValues($entity, $field_name);
```
zerolab: candidate for DI

domain/src/DomainListBuilder.php
```diff
@@ -33,27 +34,39 @@ public function getOperations(EntityInterface $entity) {
...
+    if ($admin->isForbidden()) {
+      return $operations;
     }
-    elseif (!$default) {
-      $operations['enable'] = array(
-        'title' => $this->t('Enable'),
-        'url' => Url::fromRoute('domain.inline_action', array('op' => 'enable', 'domain' => $id)),
-        'weight' => 40,
-      );
+    $account = \Drupal::currentUser();
```
zerolab: DI for currentUser() and entityTypeManager() from above?

domain/src/Tests/DomainTokenTest.php
```diff
@@ -48,7 +48,7 @@ public function testDomainTokens() {
         // context.
         $value = $domain->{$callback}();
         if ($token == '[domain:url]') {
-          $value = str_replace('user', '/', $value);
+          $value = str_replace('/user', '/', $value);
```
zerolab: 👍

domain/src/DomainAccessControlHandler.php
agentrickard: So I don't quite understand how we implement the DI pattern in a class like this. It isn't currently a service. Do we simply convert it to be one?

zerolab: something like:

```php
<?php

class DomainAccessControlHandler extends EntityAccessControlHandler implements ContainerInjectionInterface {

  /**
  * The Entity Type Manager
  *
  * @var \Drupal\Core\Entity\EntityTypeManagerInterface
  */
  protected $entityTypeManager;

  /**
   * Constructs an access control handler instance.
   *
   * @param \Drupal\Core\Entity\EntityTypeInterface $entity_type
   *   The entity type definition.
   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
   */
  public function __construct(EntityTypeInterface $entity_type, EntityTypeManagerInterface $entity_type_manager) {
    parent::__construct($entity_type);

    $this->entityTypeManager = $entity_type_manager;
  }

  /**
   * {@inheritdoc}
   */
  public static function create(ContainerInterface $container, EntityTypeInterface $entity_type) {
    return new static(
      $entity_type
      $container->get('entity_type.manager')
    );
  }


  // ...

}
```

zerolab: https://www.drupal.org/files/issues/entity_browser-access-checks-2822008-9.patch is a similar example.

agentrickard: I fumbled around with this for half an hour but can't get the create() method to fire at all, which causes fatal errors unless the EntityTypeManager defaults to NULL, which breaks the whole purpose. 

I must be missing something. Perhaps because there is a parent class that doesn't use this pattern, that method never fires? 

```
use Symfony\Component\DependencyInjection\ContainerInterface;

/**
 * Defines the access controller for the domain entity type.
 *
 * Note that this is not a node access check.
 */
class DomainAccessControlHandler extends EntityAccessControlHandler {

 /**
  * The Entity Type Manager
  *
  * @var \Drupal\Core\Entity\EntityTypeManagerInterface
  */
  protected $entityTypeManager;

  /**
   * Constructs an access control handler instance.
   *
   * @param \Drupal\Core\Entity\EntityTypeInterface $entity_type
   *   The entity type definition.
   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
   */
  public function __construct(EntityTypeInterface $entity_type, EntityTypeManagerInterface $entity_type_manager = NULL) {
    parent::__construct($entity_type);
    $this->entityTypeManager = $entity_type_manager;
  }

  /**
   * {@inheritdoc}
   */
  public static function create(ContainerInterface $container, EntityTypeInterface $entity_type) {
    return new static(
      $entity_type,
      $container->get('entity_type.manager')
    );
  }
```

agentrickard: Looks like we want the pattern from /core/modules/node/src/NodeAccessControlHandler.php, which uses createInstance().

zerolab: gah, there are so many of them. I thought I followed the `EntityAccessControlHandler`, but it might've been looking at another one. Good catch!

domain/src/DomainListBuilder.php
```diff
@@ -18,6 +27,94 @@ class DomainListBuilder extends DraggableListBuilder {
...
+   * The domain loader.
+   *
+   * @var \Drupal\domain\DomainLoaderInterface
+   */
+  protected $domainLoader;
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function createInstance(ContainerInterface $container, EntityTypeInterface $entity_type) {
```
agentrickard, line 74: This code makes me think DI is kind of silly. But I get it.

domain/src/DomainListBuilder.php
```diff
@@ -18,6 +27,94 @@ class DomainListBuilder extends DraggableListBuilder {
...
+   */
+  public function __construct(EntityTypeInterface $entity_type, EntityStorageInterface $storage, AccountInterface $account, RedirectDestinationInterface $destination_handler, EntityTypeManagerInterface $entity_type_manager, ModuleHandlerInterface $module_handler, DomainLoaderInterface $domain_loader) {
+    parent::__construct($entity_type, $storage);
+    $this->entityTypeId = $entity_type->id();
+    $this->storage = $storage;
+    $this->entityType = $entity_type;
+    $this->currentUser = $account;
+    $this->destinationHandler = $destination_handler;
+    $this->entityTypeManager = $entity_type_manager;
+    $this->accessHandler = $this->entityTypeManager->getAccessControlHandler('domain');
```
agentrickard, line 112: Should we do this in createInstance instead?

zerolab: So far I've only seen this done in __construct

agentrickard:
This latest batch of changes is based off PR #293 from @pheadeaux and it sets the pattern for handling "child" permissions for items.

In cases where items are only visible from the ability to update a specific domain, we use dual permissions in the router.

See also the change to hook_domain_operations(), which now passes an $account variable for permission checking.

Still @TODO: tests.

agentrickard:
OK. I have this one where I think it needs to be. 

This opens the gates for sites to allow different users to administer specific domains, which is very cool, especially for people who want to partition access to domain records.

It also establishes a pattern for how to grant admin access to domain functionality. See how the routes in Domain Alias are now handled.

Would love a review @zerolab

domain/domain.module
```diff
@@ -95,3 +101,113 @@ function domain_preprocess_html(array &$variables) {
...
+ * @TODO Update and abstract this function for use by other modules.
+ */
+function domain_confirm_fields($entity_type, $bundle, $text = array()) {
+  // We have reports that importing config causes this function to fail.
+
+  try {
+
+    $storage_id = $entity_type . '.' . DOMAIN_ADMIN_FIELD;
+    if (!$storage = \Drupal::entityTypeManager()->getStorage('field_storage_config')->load($storage_id)) {
+      $new_field_storage['id'] = $storage_id;
```
zerolab: Nitpick:

```php
$new_field_storage = [
 'id' => $storage_id,
  // etc..
];
```

would be cleaner. Also, $new_field_storage was not defined as an array so warnings on stricter setups.

domain/domain.module
```diff
@@ -95,3 +101,113 @@ function domain_preprocess_html(array &$variables) {
...
+      $new_field_storage['settings']['target_type'] = 'domain';
+
+      $new_field_storage = FieldStorageConfig::create($new_field_storage);
+
+      $new_field_storage->save();
+    }
+
+    $id = $storage_id = $entity_type . '.' . $bundle . '.' . DOMAIN_ADMIN_FIELD;
+    if (!$field = \Drupal::entityTypeManager()->getStorage('field_config')->load($id)) {
+      $field = array(
```
zerolab, line 152: This does what the comment above suggests.
Would be great to have them consistent.

Ditto for arrays. Core prefers short array syntax (as of 8.2.x if I am not mistaken)

domain/domain.module
```diff
@@ -95,3 +101,113 @@ function domain_preprocess_html(array &$variables) {
...
+  $form = $manager->setFormOptions($form, $form_state, DOMAIN_ADMIN_FIELD, TRUE);
+}
+
+/**
+ * Implements hook_domain_references_alter().
+ */
+function domain_domain_references_alter($query, $account, $context) {
+  // Restrict domains by assignment.
+  if ($context['field_type'] == 'admin' && $context['entity_type'] == 'user') {
+    if ($account->hasPermission('administer domains')) {
```
zerolab, line 202: Is this to prevent the case where a user has both 'administer domains' and 'assign domain administrators'?
If so, it would help to have a more explicit comment.

domain/src/DomainListBuilder.php
```diff
@@ -29,46 +126,55 @@ public function getFormId() {
...
 
-    // @TODO: permission checks.
-    if ($entity->status() && !$default) {
-      $operations['disable'] = array(
-        'title' => $this->t('Disable'),
-        'url' => Url::fromRoute('domain.inline_action', array('op' => 'disable', 'domain' => $id)),
-        'weight' => 50,
-      );
+    // If the user cannot edit domains, none of these actions are permitted.
+    $admin = $this->accessHandler->checkAccess($entity, 'update');
```
zerolab: isn't `$admin` just `$access` (or `$can_update`) in this case?
previously `$admin` was used in the context of assigned domains.
I guess more of a nitpick, than anything

domain/src/Plugin/EntityReferenceSelection/DomainAdminSelection.php
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @EntityReferenceSelection(
+ *   id = "default:domain_admin",
+ *   label = @Translation("Domain administrator selection"),
+ *   entity_types = {"domain"},
+ *   group = "default",
+ *   weight = 1
+ * )
+ */
+class DomainSelection extends DomainSelectionSelection {
```
zerolab: Should this not be `class DomainAdminSelection extends DomainSelection`?
There are no traces of `DomainSelectionSelection` other than here

zerolab (commented):
The 'default:domain_admin' selection plugin raised my eyebrows a bit. `DomainSelectionSelection` does not exist. And `DomainSelection` does not match the class file name (`DomainAdminSelection`).

Other than that a bunch of nitpicks. Testing manually now.

domain/domain.module
agentrickard: No. There are now two separate fields on the user, one for editorial access and one for domain administrative access. See the notes in hook_domain_references_alter().

domain/src/Plugin/EntityReferenceSelection/DomainAdminSelection.php
agentrickard: Dumb find/replace error.

agentrickard:
Tests are green on local. Once this passes, I'll roll alpha7.

domain/domain.module
zerolab: Ah, of course! `field_type` is provided in the Domain ER selection plugins.

agentrickard:
@zerolab I've seen this test fail on local, but cache rebuilds usually fix it.

Any ideas?

```
Fail      Other      DomainAliasListBu   34 Drupal\Tests\domain_alias\Functiona

    Drupal\Tests\domain_alias\Functional\DomainAliasListBuilderTest::testDomainListBuilder

    GuzzleHttp\Exception\ConnectException: cURL error 28: Operation timed out

    after 30001 milliseconds with 0 bytes received (see

    http://curl.haxx.se/libcurl/c/libcurl-errors.html)
```

zerolab:
Could it be a hiccup from Travis? Can you restart one those tests to double check?

agentrickard:
I restarted https://travis-ci.org/agentrickard/domain/jobs/182103628.

My suspicion is that the test runner doesn't like switching from SimpleTest to BrowserTest.

agentrickard:
Well, the concurrency setting fixed the problem. It looks like the test runner doesn't like having concurrent tests running that bridge from SimpleTest to BrowserTest.

So I suppose dropping the concurrency is the fix.

agentrickard:
Looks like separating the test runners fixed the issue!

zerolab:
Fantastic find!

agentrickard:
I had never looked at that part of the logic before. And i think Domain had 13 tests, and that's what fixed it.

PR #293, opened 2016-11-24
domain alias permissions

### LINKED ISSUE #180: Domain Alias needs proper permissions
Right now, Domain Alias is using permissions from the main module. It likely needs its own set.

tassilogroeper:
possible resolve #180
Do we need a view permission here? don't think so.


agentrickard:
This is also affected by the work in #286, which will allow for "administer aliases for assigned domains." Though that can be added later, it's good to plan for.

View permission is good form for completeness but can be handled by the base entity system so I would not bother with it.

How can we test this?

agentrickard:
I'm going to merge this and then apply it to the other issue as well.

PR #294, opened 2016-11-29
8.x 1.x config ui domain switch

davidgrayston:
#15 proof-of-concept

- Allows switching the active domain from a "sticky" form
- Saves configuration for the selected domain


zerolab:
Looks promising. A few general comments:
- DI for the `domain.negotiator` service
- Does this work with settings overrides?
- Needs tests.

**edit:** removed line about moving to modules. PEBCAK

davidgrayston:
@zerolab regarding your comments:

DI for the domain.negotiator service
- I've now done this, I hope in the correct way

Does this work with settings overrides?
- I intentionally didn't load the settings overrides as it might be confusing when editing.
- I'm now only re-applying the module overrides on non-immutable config so that the settings overrides will take effect when browsing the site, but config forms should still show module overrides.

Other issues:
- I'm currently undecided if the window should reload when switching domain.
- I think extra logic should be put in place to only apply domain module overrides on edit forms.

agentrickard:
I merged this to the *config-ui* branch for testing. First note, when enabled:

```
 ken ] : drush en domain_config_ui
The following extensions will be enabled: domain_config_ui, domain_config
Do you really want to continue? (y/n): y
Fatal error: Arrays are not allowed in class constants in /Applications/MAMP/htdocs/drupal-8/modules/domain/domain_config_ui/src/Config/Config.php on line 20
Drush command terminated abnormally due to an unrecoverable error.                                           [error]
Error: Arrays are not allowed in class constants in
/Applications/MAMP/htdocs/drupal-8/modules/domain/domain_config_ui/src/Config/Config.php, line 20
/Applications/MAMP/htdocs/drupal-8/modules/domain
```

agentrickard:
I found (or, really couldn't find) the form selector. 

The bottom right is not a natural place to find this element. That's the last place a native LTR reader would look for it.

screen shot 2016-12-06 at 4 36 50 pm


Ideally, we'd be near the Save configuration button -- if not actively changing that button.

davidgrayston:
Ah, array class constants are new in PHP 7, your fix looks better

davidgrayston:
I don't know the best way to build the domain switching into the config forms, as it's changing the global state, rather than just the config that's being saved, and it also determines the config values that will be loaded into the form. 

We can try moving the sticky switch form to a more prominent place like at the top of config forms, and reload the form to get the new config values loaded in - I'm sure this is possible with the form API.

We could also alter configuration form save buttons to read "Save for domain" when a domain has been selected.

agentrickard:
Note that we support PHP 5.5, 5.6 and 7.  (I'm using 5.6.10.)

See https://travis-ci.org/agentrickard/domain/builds/181052317 and https://www.drupal.org/docs/7/system-requirements/php#php_required

domain_config/src/DomainConfigOverrider.php
```diff
@@ -176,7 +176,15 @@ protected function initiateContext() {
     $this->languageManager = \Drupal::languageManager();
     $this->language = $this->languageManager->getCurrentLanguage();
     // Get the domain context.
-    $this->domain = $this->domainNegotiator->getActiveDomain(TRUE);
+    $this->setDomain($this->domainNegotiator->getActiveDomain(TRUE));
```
davidgrayston: I had to introduce this setter to allow the domain to be switched on AJAX requests

domain_config/src/DomainConfigOverrider.php
```diff
@@ -176,7 +176,15 @@ protected function initiateContext() {
...
   }
 
+  /**
```
davidgrayston: I had to introduce this setter to allow the domain to be switched on AJAX requests

domain_config_ui/domain_config_ui.services.yml
```diff
@@ -0,0 +1,23 @@
+services:
+  domain_config_ui.negotiator:
+    class: Drupal\domain_config_ui\DomainConfigUINegotiator
+    tags:
+      - { name: persist }
+    arguments: ['@request_stack', '@module_handler', '@domain.loader', '@config.factory']
+    calls:
+      - [setDomainConfigOverrider, ['@domain_config.overrider']]
```
davidgrayston: Passing the domain overrider to the UI negotiator so that the domain can be switched on AJAX requests

domain_config_ui/src/Config/Config.php
```diff
@@ -0,0 +1,104 @@
...
+namespace Drupal\domain_config_ui\Config;
+
+use Drupal\Core\Config\Config as CoreConfig;
+use Drupal\domain\DomainNegotiatorInterface;
+
+/**
+ * Extend core Config class to save domain specific configuration.
+ */
+class Config extends CoreConfig {
+  /**
```
davidgrayston: Changed to allowed list as it's proving difficult to make this work on every form.

domain_config_ui/domain_config_ui.module
```diff
@@ -0,0 +1,94 @@
...
+  }
+}
+
+/**
+ * Checks if provided form can be used to save domain specic configuration.
+ *
+ * @param array $form
+ * @return boolean
+ */
+function domain_config_ui_form_is_allowed($form) {
```
davidgrayston: Added allowed form list to prevent domain switch appearing on every admin form

domain_config/src/DomainConfigOverrider.php
```diff
@@ -176,7 +176,15 @@ protected function initiateContext() {
...
+   * Set the domain context.
+   *
+   * @param DomainInterface $domain
+   */
+  public function setDomain(DomainInterface $domain) {
+    $this->domain = $domain;
+  }
```
davidgrayston: I had to introduce this setter to allow the domain to be switched on AJAX requests

domain_config_ui/domain_config_ui.module
```diff
@@ -0,0 +1,94 @@
...
+  $allowed = array(
+    'system_site_information_settings',
+    'system_themes_admin_form',
+  );
```
davidgrayston: Added allowed form list to prevent domain switch appearing on every admin form

domain_config_ui/src/Config/Config.php
```diff
@@ -0,0 +1,104 @@
...
+   * List of config that can be saved for a specific domain.
+   * Use * for wildcards.
+   */
+  protected $allowedConfig = [
+    'system.site',
+    'system.theme*',
+  ];
```
davidgrayston: Changed to allowed list as it's proving difficult to make this work on every form.

domain_config_ui/src/DomainConfigUINegotiator.php
```diff
@@ -0,0 +1,76 @@
...
+   *
+   * @param string $domain_id
+   */
+  public function setSelectedDomain($domain_id) {
+    if ($domain = $this->domainLoader->load($domain_id)) {
+      // Set session for subsequent request.
+      $_SESSION['domain_config_ui']['config_save_domain'] = $domain_id;
+      // Switch active domain now so that selected domain configuration can be loaded immediatly.
+      // This is primarily for switching domain with AJAX request.
+      $this->domainConfigOverrider->setDomain($domain);
```
davidgrayston: Setting the overrider domain so that AJAX requests can return selected domain configuration

domain_config_ui/src/DomainConfigUINegotiator.php
```diff
@@ -0,0 +1,76 @@
...
+   */
+  protected function negotiateActiveDomain() {
+    // Set http host to be that of the selected domain to configure.
+    if ($selected_domain = $this->getSelectedDomain()) {
+      $httpHost = $selected_domain->getHostname();
+    }
+    else {
+      $httpHost = $this->negotiateActiveHostname();
+    }
+    $this->setRequestDomain($httpHost);
```
davidgrayston: I don't think spoofing the request domain to load is the correct solution - will try to move this to a better place

domain_config_ui/domain_config_ui.module
```diff
@@ -1,19 +1,79 @@
...
+  }
+  $form['config_save_domain'] = array(
+    '#type' => 'select',
+    '#weight' => -10,
+    '#title' => 'Domain',
+    '#options' => array_merge(['' => 'All Domains'], \Drupal::service('domain.loader')->loadOptionsList()),
+    '#default_value' => $selected,
+    '#ajax' => array(
+      'callback' => 'domain_config_ui_domain_switch_form_callback',
+      ''
```
zerolab: Should this empty string be here?

domain_config_ui/domain_config_ui.module
```diff
@@ -1,19 +1,79 @@
...
+  else {
+    $selected = $form_state->getValue('config_save_domain');
+  }
+  $form['config_save_domain'] = array(
+    '#type' => 'select',
+    '#weight' => -10,
+    '#title' => 'Domain',
+    '#options' => array_merge(['' => 'All Domains'], \Drupal::service('domain.loader')->loadOptionsList()),
+    '#default_value' => $selected,
+    '#ajax' => array(
```
zerolab: Let's switch to array shorthand everywhere.

domain_config_ui/domain_config_ui.module
```diff
@@ -1,19 +1,79 @@
...
-  if ($is_admin) {
-    $form = \Drupal::formBuilder()->getForm('Drupal\domain_config_ui\Form\SwitchForm');
-    $variables['page']['content']['domain_config_ui_switch'] = $form;
+  // Add a message below the form to remind the administrator which domain they are currently configuring.
+  if ($selected_domain) {
+    $message = new TranslatableMarkup('Configuration will be saved for @domain', array(
+      '@domain' => $selected_domain->label(),
+    ));
+    $form['domain_message'] = array(
+      '#markup' => new FormattableMarkup('<div class="messages messages--warning">@message</div>', array(
```
zerolab:
```
$form['domain_message'] = [
  '#type' => 'container',
  '#markup' => $message,
  '#attributes' => ['class' => ['messages', 'messages--warning']],
  '#weight' => 1000,
];
```

is much cleaner

zerolab:
This is starting to look nice.

domain_config_ui/domain_config_ui.services.yml
```diff
@@ -0,0 +1,16 @@
+services:
+  domain_config_ui.manager:
+    class: Drupal\domain_config_ui\DomainConfigUIManager
+    arguments: ['@config.storage', '@domain.loader', '@language_manager']
```
davidgrayston, line 4: I'm no longer extending the domain config overrider or negotiator. DomainConfigUIManager service will now manage the language/domain switching.

domain_config_ui/domain_config_ui.services.yml
```diff
@@ -0,0 +1,16 @@
...
+
+  domain_config_ui.factory:
+    class: Drupal\domain_config_ui\Config\ConfigFactory
+    tags:
+      - { name: event_subscriber }
+      - { name: service_collector, tag: 'config.factory.override', call: addOverride }
+    arguments: ['@config.storage', '@event_dispatcher', '@config.typed']
+    calls:
+      - [setDomainConfigUIManager, ['@domain_config_ui.manager']]
```
davidgrayston, line 13: Config UI manager is being passed to the config factory so that it can be used to load up selected domain/language configuration.

domain_config_ui/src/DomainConfigUIManager.php
```diff
@@ -0,0 +1,184 @@
...
+    // Get the domain context.
+    if ($domain = $this->getSelectedDomain()) {
+      $this->domain = $domain;
+    }
+  }
+
+  /**
+   * Load only overrides for selected domain and language.
+   */
+  public function loadOverrides($names) {
```
davidgrayston, line 77: Config UI manager loads overrides for the current selected domain/language

domain_config_ui/domain_config_ui.module
```diff
@@ -0,0 +1,110 @@
...
+    '#type' => 'select',
+    '#title' => 'Domain',
+    '#options' => array_merge(['' => 'All Domains'], \Drupal::service('domain.loader')->loadOptionsList()),
+    '#default_value' => $selected_domain ? $selected_domain->id() : '',
+    '#ajax' => [
+      'callback' => 'domain_config_ui_domain_switch_form_callback',
+    ],
+  ];
+
+  // Add language select field.
```
davidgrayston, line 66: Language can now be selected

domain_config_ui/domain_config_ui.module
davidgrayston: This was a typo, now cleaned up

domain_config_ui/src/Form/SwitchForm.php
```diff
@@ -0,0 +1,113 @@
+<?php
```
davidgrayston, line 1: I've re-introduced the switch form for use on pages that aren't conventional config forms, such as the theme settings page. This form will reload the page when the domain is switched.

domain_config_ui/src/Config/ConfigFactory.php
```diff
@@ -0,0 +1,160 @@
...
+      if (!empty($storage_data)) {
+        // Only get domain overrides if we have configuration to override.
+        $module_overrides = $this->loadDomainOverrides($names);
+      }
+
+      foreach ($storage_data as $name => $data) {
+        $cache_key = $this->getConfigCacheKey($name, $immutable);
+
+        if (isset($module_overrides[$name])) {
+          $this->cache[$cache_key]->setModuleOverride($module_overrides[$name]);
```
davidgrayston, line 107: I've applied the domain overrides on immutable config so that unconventional admin pages (like theme settings) show the domain domain specific config

davidgrayston:
I still need to go through the known paths/forms/config that should be added to the allowed list by default.

The `/admin/config/system/site-information` form seems to be working fine, but the theme pages are still a bit clunky to use at the moment.

domain_config_ui/domain_config_ui.services.yml
agentrickard: Was there a specific code reason that made doing so necessary? 

Ideally, we'd handle all of the negotiation inside the DomainConfigOverrider, or, if needed, create a DomainConfigManager service for this.

My thought here is that the UI piece should not handle any of the negotiation logic, just the form handling, so that it could be swapped.

davidgrayston: The reason was to keep the loading of config that's being edited as self-contained as possible.

Spoofing the requested domain for the negotiator meant that all the config on the page also got swapped. It didn't seem right to have an admin tool overriding the negotiator and change its behaviour.

I'm also not sure if we want setting overrides to be applied to editable values?

I can change this back if you think swapping all config on the page is better?

davidgrayston: On second thoughts, I am swapping all allowed config with the manager so it might not be helping much.

I can change back to extending the negotiator as before and see what problems I encounter.

davidgrayston: I've changed it back to use the negotiator on a new branch here https://github.com/davidgrayston/domain/tree/8.x-1.x-config_ui_domain_switch_negotiator

Diff with current pull request branch: https://github.com/davidgrayston/domain/compare/8.x-1.x-config_ui_domain_switch...8.x-1.x-config_ui_domain_switch_negotiator

I had to make some changes to DomainConfigOverrider to make this work:
- Setters to allow the domain and language to be set
- Changed protected getDomainConfigName method to be public

This doesn't seem like the correct solution in its current state

agentrickard: I haven't had a chance to really dive into this code. The part that bothers me is the introduction of config handling to the DomainNegotiator. That class/service should not really be Config aware. 

I have no problem making getDomainConfigName public, but I haven't worked through the pro-con.

In general:

* The core Domain module and its classes should handle domain registration and negotiation. That's all.
* Domain Config should handle storage and retrieval of config overrides.
* Domain Config UI should put a form in from of the read/write aspects of Domain Config.

Put another way, we should be able to programatically write an override through Domain Config.

davidgrayston: Due to the saving of configuration not being tied to forms in a standardised way, and there being no definitive way of knowing which forms/pages handle config, it's quite tricky to implement this.

I don't think there's a neat way of loading the config overrides only for forms values (and admin pages), so we have to swap the config on the entire page - it's essentially swapping the domain being viewed. This is why I initially made a global domain switcher (sticky form), to make it clear that you were changing the selected domain globally.

Now that I'm altering "known" config forms, the user will not know which domain is currently selected until they go to an admin form with the switcher field. Perhaps we should go back the the domain switch implementation but make the form much more prominent (e.g. a fixed strap at the top of the page)?

It might be quicker if you take these ideas as a proof of concept and implement as you have described above?

agentrickard: I am working on finishing other parts of the alpha/beta release, so speed isn't something I can provide.

I need to take a deeper look at what you've done and the problems you define and try to give some better direction.

Using a config_ui_manager to handle forms makes sense. I just don't want that to bleed over into handling save/load.

andypost: That's really tricky place, d.o core has a lot of "config override bleeding" issues fixed, also working on  domain_lang module it became clear that priority of loading overrides highly coupled to language manager initialization.
Also some overrides can be done via settings.php and somehow that should affect in UI

andypost:
Here's addition to show widget with permission only https://github.com/skilld-labs/domain/pull/1

dmiric:
For the last few days before I saw this I was working on the solving this problem on my own. I'll try to describe my approach to this.

So what I did was:

- Introduce a hook in config save (hook_config_alter) that accepts config name and data
- Overridden config factory method getEditable to return overridden configuration values for the domain I'm currently on
- And then In hook_config_alter I would simply change config that is being saved to have the name and data I wanted to save for that site

So far everything is working pretty well. I can edit all simple configuration blocks and views its really early to tell if I can achive everything I want this way but looks promising.

Of course I'm also not sure If Drupal community would accept to have that hook in config save. 

What do you think about that approach?

If it was not clear enough. You basically visit the site from domain you want to change config and make changes in usual places as if you were building a normal Drupal site.

dmiric:
I'm also thinking about making a domain template so you could inherit configuration from template and then introduce minor changes on domain it self so we don't have so many config files.

andypost:
@dmiric Use events instead of hooks for config, see `ConfigEvents` core class
BTW we using this branch on a project and we are going to add language & number of overrides to the floating block, also looks toolbar is a proper place for that

dmiric:
I tried to do it using config::save event, but I could not figure out how
to avoid saving the base site configuration. Would you happen to know how I
could do that?

andypost:
After trial on this UI following bugs found

- domain override always picked from session (set by floating box), probably it needs to use query string argument to pass, so UI needs to redirect to the same page but appending URL with `?domain=ID`
- edit forms always using original config, overriden config factory probably should care to load override if exists
- language in name of override makes a lot of overrides and not using config translation

andypost:
One more core change in this area https://www.drupal.org/node/2930829

agentrickard:
Hooray! That change was a long time coming.

rivimey:
I created a new PR for this as #426 which (apart from unrelated 8.6 issues) merges and builds successfully.

andypost:
Actually instead of 2 PRs better to create repo like we used to hack https://github.com/skilld-labs/domain_config_ui and re-do it

Core in progress of CMI 2.0 so that may affect domain_config itself (I still looking to move domain overrides to own collections instead of prefixing)

rivimey:
@andypost Apologies, but I don't really understand "create repo like we used to hack...and re-do": I'm still finding my way around here.

[My intent here was only to use travis to check the patch again, given it has been a while since it was pushed. I intend to close the PR. ]

agentrickard:
I'm going to pull https://github.com/skilld-labs/domain_config_ui into a new branch here for testing.

davidgrayston:
Happy to apply the changes to the branch on this PR if that helps?

agentrickard:
Well, I moved things to #434 so that it would be part of the main repo. So new changes would go there.

PR #296, opened 2016-12-05
Proposed solution for unique domain id.

jurcello:


agentrickard:
We're still in alpha. Update hooks are not supported.

I appreciate the PR, but this was submitted without any supporting context.

Please:

* Reference an open issue 
* Describe the changes

zerolab:
I'd like to add that I believe this is trying to solve a non-issue here.

`createDomainId()` is called when the new domain is created and the hostname is a unique enough identifier to use as a seed for crc32.

If there is a real issue, this needs tests demonstrating it.

jurcello:
As I said in the drupal.org issue (https://www.drupal.org/node/2833053), this not an non-issue:
if the country path module is used, the hostname is the same for more than one domain.

I need to distinguish domains based on a path prefix. If the id is based on the hostname, the whole domain module fails in that case. That means that the domain module should be patched from then on.
In my opinion.

I will also file an issue and create a new pull request without the update hook.

zerolab:
Fair enough. With the new issue/d.o. issue there is enough context.

agentrickard:
@jurcello Thank you for the clarification.

In future, please link to d.o. issues when filing PRs. I don't check d.o. frequently.

I was not aware of country_path, but I would think the burden of handling this ID case falls to that module, not this one.

However, since we know that the machine name must be unique, then this solution does seem simpler.

agentrickard:
Note: In this case, I think the alpha-to-alpha update hook makes sense, since we are doing a fundamental alteration of a critical piece of data, which might affect people like @zerolab who are running the module in production.

Can we put in some manner of check so that this update only runs if required? We haven't written the "real" upgrade path yet, and I don't want this to fire when it's not needed.

Perhaps we can save a state variable? Or check the version id of the module?

zerolab:
#298 replaces this PR and does not have the update hook.

I'd rather the update hook contents were added to the changelog/ upgrade considerations to be run manually by developers if they want to.

agentrickard:
So you're suggesting we do what? Put a note in the CHANGELOG that says "If you're running a release prior to Alpha X, grab this code and run a custom update hook?"

zerolab:
Pretty much. "Run this code, if you need to".
The point is, where it works there is no need to change the IDs. So this change can safely go in without the update hook.

agentrickard:
I've moved this to #300.

Issue #297, opened 2016-12-06
Domain id is based on hostname, which is not always unique


jurcello:
The domain id is now based on the hostname, which is not unique in all cases (for instance when using country_path):

public function createDomainId() {
    // We cannot reliably use sequences (1, 2, 3) because those can be different
    // across environments. Instead, we use the crc32 hash function to create a
    // unique numeric id for each domain. In some systems (Windows?) we have
    // reports of crc32 returning a negative number. Issue #2794047.
    $this->domain_id = abs((int) crc32($this->getHostname()));
  }

There are two possible ways to distinguish between domains: hostname or path prefix.
Although path prefix is suboptimal, there are use cases where it is needed.
When the domain id is based on the hostname this does not support path prefixes.
When the domain id is based on the id (which by definition should be unique) this problem is solved.

agentrickard:
I've moved this work into #300. Thanks for the help!

PR #298, opened 2016-12-06
Other id in order to support prefix domain matching.

jurcello:
This is a new pull request for issue #297 and drupal.org issue https://www.drupal.org/node/2833053.
I hope it's more clear now what the problem is.


agentrickard:
Yeah. the problem is people are writing new modules and not telling me :-)

agentrickard:
Based on the discussion, I've moved this to #300, which could use a review.

PR #300, opened 2016-12-06
Alpha update

agentrickard:
Takes the work in #296 and #298 and provides a reliable upgrade path.

I think this is worth doing, even though we don't support all alpha changes (especially permissions.) This code changes underlying data structures and poses a security issue, so providing a means to run the update seems responsible.

### DIFF
--- a/domain_alpha/domain_alpha.install
+++ b/domain_alpha/domain_alpha.install
@@ -8,14 +8,23 @@
 /**
  * Update domain id to new value.
  */
-function domain_alpha_update_8000(&$sandbox) {
+function domain_alpha_update_8001(&$sandbox) {
   // Load all domains and update the id.
+  $rebuld = FALSE;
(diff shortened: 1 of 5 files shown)


domain_alpha/domain_alpha.install
```diff
@@ -0,0 +1,30 @@
...
+ * @file
+ * Pre-release update hooks for Domain modules.
+ */
+
+/**
+ * Update domain id to new value.
+ */
+function domain_alpha_update_8001(&$sandbox) {
+  // Load all domains and update the id.
+  $rebuld = FALSE;
```
zerolab: `s/$rebuld/$rebuild/`

zerolab:
one small typo in the update hook. LGTM otherwise

agentrickard:
That's non a meaningless typo ;-)

PR #302, opened 2016-12-14
Fixing saving of domain settings when config has not yet been created

davidgrayston:
I noticed that domain settings cannot be saved if the configuration has not yet been created.

This is because a `$this->config()` returns a new config object on each iteration of the foreach loop in the submit handler.

To reproduce the issue, remove domain.settings.yml from the sync directory and re-import config. The settings can no longer be saved through the form.

I've updated the code to get the config object once, then reuse in the foreach loop.


zerolab:
LGMT!

Issue #311, opened 2017-01-10
New users can see the Domain Administrator field


agentrickard:
When signing up for an account at /user/register, all options of the Domain Administrator field are shown.

This is bad :-)

agentrickard:
In my tests, the field handler is always getting set to 'default:domain', when it should be 'default:domain_admin'.

I don't understand why, though I did just start using PHP 7.1 locally.

Anyone else seeing this?

agentrickard:
It's an issue with calling domain_cofirm_fields() during hook_install().

agentrickard:
It looks like Drupal is not recognizing the DomainAdminSelection plugin. It should be available as a field configuration option.

Issue #310, opened 2017-01-05
Update issue in Domain Alpha module


nedjo:
The Domain Alpha module contains a (much appreciated!) update needed for sites updating from a previous release. However, since this module is not yet installed on those sites, the update won't run. (On install, the `domain_alpha` schema value will be set to that of the update.) Needed:

* Copy the code from the `domain_alpha_update_8001()` update into `hook_install()`.

A workaround is:

* Enable Domain Alpha.
* Enable Devel.
* Paste in content of `domain_alpha_update_8001()` from `domain_alpha.info.yml` to form at /devel/php and submit.

agentrickard:
Sigh. Good point.

PR #312, opened 2017-01-11
Try to fix the field install of the admin field.

agentrickard:
Let's run the testbot on this. (See #311

### DIFF
--- a/domain/config/install/field.field.user.user.field_domain_admin.yml
+++ b/domain/config/install/field.field.user.user.field_domain_admin.yml
@@ -16,7 +16,7 @@ translatable: false
 default_value: {  }
 default_value_callback: ''
 settings:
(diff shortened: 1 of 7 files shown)


agentrickard:
This really needs a test of the user/register form.

andypost:
I think better to change field storage in domain_access `persist_with_no_fields: true` and add `enforce` dependency to make them uninstallable automatically when doamin-access uninstalled

agentrickard:
That's a separate issue. This issue is that the wrong EntityReferneceSelect plugin was being assigned to the domain_admin field.

agentrickard:
Added another test to ensure that edit and admin fields are handled differently. This should be ready to go once it passes.

We'll then update the field definition as part of #310

PR #314, opened 2017-01-11
Claryfy variable naming in active domain subscriber

andypost:
Loking at `\Drupal\domain\EventSubscriber\DomainSubscriber::onKernelRequestDomain()` it somehow not clear what the logic behinde the redirect to active domain

### DIFF
--- a/domain/src/EventSubscriber/DomainSubscriber.php
+++ b/domain/src/EventSubscriber/DomainSubscriber.php
@@ -74,22 +74,20 @@ class DomainSubscriber implements EventSubscriberInterface {
    *
    * @param \Symfony\Component\HttpKernel\Event\GetResponseEvent $event
    *   The Event to process.
+   *
+   * @see domain_alias_domain_request_alter
    */
   public function onKernelRequestDomain(GetResponseEvent $event) {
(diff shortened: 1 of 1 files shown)


domain/src/EventSubscriber/DomainSubscriber.php
```diff
@@ -74,22 +74,20 @@ public function __construct(DomainNegotiatorInterface $negotiator, DomainLoaderI
...
+   * @see domain_alias_domain_request_alter
    */
   public function onKernelRequestDomain(GetResponseEvent $event) {
-    $redirect = FALSE;
     // Negotiate the request and set domain context.
     /** @var \Drupal\domain\DomainInterface $domain */
     if ($domain = $this->domainNegotiator->getActiveDomain(TRUE)) {
       $domain_url = $domain->getUrl();
       if ($domain_url) {
-        $redirect_type = $domain->getRedirect();
```
agentrickard, line 85: The reasoning on name here is because the redirect could be different, though I think status is perfectly clear.

The only code that sets a redirect for a domain is in domain_alias_domain_request_alter().

PR #319, opened 2017-01-17
Show disallowed options.

agentrickard:
This one could use some review and opinion. In some cases, this feature is desirable. Perhaps not in others.

I don't really want to make another permission here, but could.

### DIFF
--- a/domain/src/DomainElementManager.php
+++ b/domain/src/DomainElementManager.php
@@ -3,6 +3,7 @@
 namespace Drupal\domain;
 
 use Drupal\Core\Form\FormStateInterface;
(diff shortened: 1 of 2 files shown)


zerolab:
Failing with:

```
Fatal error: Uncaught Error: Call to undefined method 
Drupal\domain_access\Tests\DomainAccessFieldTest::assertNoRawText() in 
/home/travis/build/agentrickard/domain/domain_access/src/Tests/DomainAccessFieldTest.php:72
```

agentrickard:
Ok, now that it works, the question is "Does it need a permission or a setting to enable/disable for some users"?

agentrickard:
No response here, so let's test and see what people think.

PR #320, opened 2017-01-30
Dependencies are wrong https://www.drupal.org/node/2848014

andypost:
Issue #2848014 by marcelovani, andypost: Invalid dependency array on update hook for field storage of user.field_domain_admin

Entity reference fields provides proper permissions by default

### DIFF
--- a/domain/domain.install
+++ b/domain/domain.install
@@ -33,7 +33,7 @@ function domain_uninstall() {
   $field_storage_config = \Drupal::entityTypeManager()
     ->getStorage('field_storage_config');
 
-  $id = $type . '.' . DOMAIN_ADMIN_FIELD;
+  $id = 'user.' . DOMAIN_ADMIN_FIELD;
   if ($field = $field_storage_config->load($id)) {
     $field->delete();
   }
--- a/domain/domain.module
+++ b/domain/domain.module
@@ -138,7 +138,6 @@ function domain_confirm_fields($entity_type, $bundle, $handler = 'default:domain
         'id' => $storage_id,
         'field_name' => DOMAIN_ADMIN_FIELD,
         'type' => 'entity_reference',
-        'dependencies' => 'domain' , 'user',
         'entity_type' => 'user',
         'cardinality' => -1,
         'module' => 'entity_reference',


zerolab:
This is a domain specific field, so arguably if domain is uninstalled, the field should go too. https://www.drupal.org/node/2848014#comment-11899423 feels like the right approach

andypost:
patch is wrong and default values are

```
dependencies:
  module:
    - domain
    - user
```

I also added proper uninstall

agentrickard:
So I see two errors here -- one a simple syntax goof with the dependencies array. The other in the uninstall function.

I think the field storage needs that dependency, doesn't it?

Do we also need to delete the field instance on uninstall?

I'll dig into this more over the weekend.

andypost:
Uninstall already deletes field storage so there should be proper field name.
Dependencies for field storage is not needed because it deleted in hook_uninstall()
But if you wanna to allow core to delete field for you you need to provide `enforced` dependency for field storage 

The field (instance in terms of d7) already has this dependency because referring to domain entities

agentrickard:
Right, but this function is just replicating what's in the config export (in the install YAML).

```
dependencies:
  module:
    - domain
    - user
```

When you add a field via the UI and export it, it adds the entity the field is attached to and, in this case, the entity providing the reference.

I would keep the dependencies to be consistent. 

If you know, by the way, how to install straight from the YAML, that would be even better.

andypost:
I think better to install from yaml like 
- http://cgit.drupalcode.org/password_policy/tree/config/install/field.storage.user.field_last_password_reset.yml?h=8.x-3.x#n7
- http://cgit.drupalcode.org/password_policy/tree/config/install/field.field.user.user.field_last_password_reset.yml?h=8.x-3.x#n9

I will update MR tomorrow with this changes

agentrickard:
If we can do that, we can get rid of the very ugly confim_fields() tricks that I have been using. With Domain Access, we need to install fields dynamically based on content types, which is where this function started. I wrote it years ago and haven't replaced it with better code.

agentrickard:
I went with the simple array fix here, let's fix the install in a followup.

Issue #321, opened 2017-02-05
Better field installation


agentrickard:
If we can, we should replace domain_confirm_fields() and domain_access_confirm_fields() with dynamic field installation from the field YML files.

See https://www.drupal.org/node/2849888

agentrickard:
Fixed.

PR #327, opened 2017-02-07
Use config instead of hook_uninstall()

andypost:
This is a follow-up to https://www.drupal.org/node/2848014#comment-11917448

Fields are (un)installed from shipped config that makes hooks obsolete


andypost:
Somehow tests are failed, ref #321

andypost:
tests fail because field is not added to user form display
```
-    entity_get_form_display($entity_type, $bundle, 'default')
 -      ->setComponent(DOMAIN_ADMIN_FIELD, array(
 -        'type' => 'options_buttons',
 -        'weight' => 50,
 -      ))
 -      ->save();
```

andypost:
Now uninstall tells what will be uninstalled

config-uninstall

agentrickard:
Two questions:

1) Is this testable, or can we consider it covered by core tests?

2) How does this affect (or not) the need for dynamic field creation for nodes?

andypost:
1) config dependencies have good testing coverage in core, that's why I suggest to rely on them, see http://cgit.drupalcode.org/drupal/tree/core/modules/config/src/Tests/ConfigOtherModuleTest.php as example for dependency on modules
2) dynamic field creation is not affected because fields depends on their storage that now have enforced dependencies on modules that provides them

basically when module will be uninstalled all dependent field storages will cause fields to be deleted - that's what I trying to show with screenshot

agentrickard:
Got it. This is great.

I don't expect that these changes would apply to anyone who has installed the alpha versions of the module, would they?

I don't think we need to provide an update for that, though, since it should be clear what people need to do.

andypost:
there's `domain_update_8001` already so it's easy to add update hooks
but I think no reason for alpha stage

agentrickard:
I agree. Thanks again for this. Since tests pass, I'm going to commit and cleanup if necessary.

Issue #147, opened 2016-01-22
Dependency Injection in DomainValidator


agentrickard:
Use proper DI in DomainValidator.

agentrickard:
The EntityTypeManager and ModuleHandler services are not called correctly.

We should merge this with #11 -- which is about separating validation steps, perhaps into Plugins.

PR #332, opened 2017-02-12
Refactor domain validation

### LINKED ISSUE #11: Separate the validation
Take the DomainValidator::validate() method and break it into pieces, possibly as Plugins, for easier testing.

andypost:
- change interface to pass hostname instead of domain entity
- inject dependencies
- fix drush domain-add
- clean-up doc blocks

Closes #11 & #147

### DIFF
--- a/domain/src/DomainForm.php
+++ b/domain/src/DomainForm.php
@@ -5,6 +5,7 @@ namespace Drupal\domain;
 use Drupal\Core\Entity\EntityForm;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Entity\EntityStorageInterface;
+use Drupal\Core\Render\RendererInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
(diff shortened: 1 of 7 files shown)


andypost:
Failed test shows that domain record needs kinda validation for existing hostname at entity level validator

andypost:
Filed #333
Config entities using preSave() and exceptions to validate ability to save

andypost:
> berdir> andypost: I would just add standard form api validation. throwing exceptions is IMHO only for the kind of invalid data that will result exceptions/broken drupal later, e.g. invalid/duplicate field names and so on

agentrickard:
This looks pretty good. I didn't know you could do try/catch inside the tests like that.

domain/src/DomainForm.php
```diff
@@ -120,36 +155,40 @@ public function form(array $form, FormStateInterface $form_state) {
...
+        '#items' => $errors,
+      ];
+      $message = $this->renderer->renderPlain($message);
+      $form_state->setErrorByName('hostname', $message);
+    }
+    // Validate if the same hostname exists.
+    // Do not use domain loader because it may change hostname.
+    $existing = $this->storage->loadByProperties(['hostname' => $hostname]);
+    $existing = reset($existing);
+    if ($existing && $existing->id() != $entity->id()) {
```
andypost, line 176: maybe move this 3 lines to "validator" method like `validateExistingHostname($hostname, $id)` but it looks useless
But probably #11 supposed to do that

agentrickard: The idea of #11 was to split each validation step into a separate, testable method. Think about validation methods as abstract elements that could be reused by other applications. 

I don't know if we have to check for id() uniqueness -- or if core will do it for us.

If we have to do it, it would stay with the form validator, because it's an entity validation issue, not a valid hostname issue.

andypost: The intend here is to prevent to add another domain with same hostname, also it prevents edit to change hostname for existing

BTW I suppose entity level validation to properly handle REST operations on domain records

agentrickard: You should be able to edit existing hostnames. Sometimes people commit typos in names. Often, people change out prod/dev domains as canonical.

andypost: Yep, I mean the case when user trying to edit hostname but there's other record with the same one
Looks that needs extra test to show

andypost:
BTW validation of hostname does not have dependencies, so no reason to have it as service
Basic helper class is enough

PR #335, opened 2017-02-13
Make all affiliates default value configurable

### LINKED ISSUE #156: Make all affiliates default value configurable?
Ideally, this value would be configurable per content type. Not sure if that is possible. Or if we just let field module handle things.

andypost:
Closes #156 

PS: Also removed useless calls to container


agentrickard:
I think this one needs an update hook in domain_alpha, since we would break existing installs by removing the callback function.

Alternately, we just don't remove the function.

andypost:
Added `hook_update` also removed default value setting because for boolean items it always false

andypost:
Updated changelog, looks now ready

PR #336, opened 2017-02-15
Review Domain entity & remove DomainViewBuilder

### LINKED ISSUE #148: DomainViewBuilder review
There are some oddities noted in DomainViewBuilder. Investigate.

andypost:
Closes #148 

- Removes `DomainEntityBuilder` because it suppose only for "viewable" entities
- Use `count()` in `preCreate()` instead of loading entities
- `status` & `uuid` properties are defined in base class so should not be duplicated
- `status` annotation should be set for config entities that allowed enable/disable- Clean-up doc-blocks to conform current standards (domain & interface now passes codereview)
- Removes `isDefaultRevision` method because it defined for content entities `\Drupal\Core\Entity\RevisionableInterface::isDefaultRevision()`

**TODO** looks `getRawPath()` & `buildUrl()` methods are defined but never used, makes sense to remove (they are not a part of interface)


agentrickard:
buildURL() is used in DomainAccessField.php, probably as a shim for a case where I couldn't figure out how to generate a proper link without a route. (And to prevent, I think, the form from linking to the domain entity.)

In D7 we used similar functions to build absolute links, but they likely aren't needed now.

domain/src/Entity/Domain.php
```diff
@@ -475,11 +456,4 @@ public function getMatchType() {
     return $this->matchType;
   }
 
- /**
-  * This is being fired for some reason if the values are shown on a
```
agentrickard, line 479: This was originally here because (I think) of a bug in core.

andypost:
Thanx for pointer, will dig about link generation

agentrickard:
It's worth noting that the DomainLoader class was created because Drupal 8.0 wasn't quite stable from an entity API pov. There were methods that I wanted that were missing, and I wanted something stable to develop against.

It is possible that we might be able to remove the Loader (and the DomainCreator) now that core is much more advanced.

I think we're fine keeping both but wanted to note that. I have no real interest in rewriting to remove those classes.

domain/src/Entity/Domain.php
andypost: Probably that was in times befor interfaces for config and content got split

agentrickard: I suspect you'll see a lot of that in the code cleanup.

andypost:
I will care to cleanup and provide BC shims for loader&creator that's why I started to review & fix file by file

agentrickard:
I'm a huge fan of creating convenience methods and functions for other developers, which is why we have things like DomainLoader::loadDefaultDomain().

Those might be something that can be moved to a more standard storage extension, but I would like to keep them.

andypost:
Sure, it's a part of interface that's why BC should be kept

agentrickard:
For all my ignorance of OOP, i was very diligent about creating interfaces. :-)

andypost:
Yep, they are perfectly architectured!
So it's easy to navigate and polish implementation

agentrickard:
I blame @crell for that. :-)

Crell:
Blame accepted!

PR #341, opened 2017-03-12
Implement hook_requirements for Domain Content.

agentrickard:
See https://www.drupal.org/node/2827167

### DIFF
--- a/domain_content/domain_content.module
+++ b/domain_content/domain_content.module
@@ -33,3 +33,37 @@ function domain_content_domain_operations(DomainInterface $domain, AccountInterf
 
   return $operations;
 }
+
+/**
+ * Implements hook_requirements.
+ */
+function domain_content_requirements($phase) {
+  $requirements = array();
+  $allow = TRUE;
+  if ($phase == 'install') {
+    $list['user'] = 'user';
+    $node_types = \Drupal::entityTypeManager()->getStorage('node_type')->loadMultiple();
+    foreach ($node_types as $type => $info) {
+      $list[$type] = 'node';
+    }
+    // Check for required fields.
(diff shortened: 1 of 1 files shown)


domain_content/domain_content.module
```diff
@@ -33,3 +33,37 @@ function domain_content_domain_operations(DomainInterface $domain, AccountInterf
...
+    $list['user'] = 'user';
+    $node_types = \Drupal::entityTypeManager()->getStorage('node_type')->loadMultiple();
+    foreach ($node_types as $type => $info) {
+      $list[$type] = 'node';
+    }
+    // Check for required fields.
+    foreach ($list as $bundle => $entity_type) {
+      $id = $entity_type . '.' . $bundle . '.' . DOMAIN_ACCESS_FIELD;
+      if (!$field = \Drupal::entityTypeManager()->getStorage('field_config')->load($id)) {
+        $allow = FALSE;
```
zerolab, line 53: Maybe break early? if there is at least one FALSE, there is no need to check all of them

domain_content/domain_content.module
```diff
@@ -33,3 +33,37 @@ function domain_content_domain_operations(DomainInterface $domain, AccountInterf
 
   return $operations;
 }
+
+/**
+ * Implements hook_requirements.
+ */
+function domain_content_requirements($phase) {
+  $requirements = array();
```
zerolab: Nitpick: use array short-form. Especially that core has moved to that - https://groups.drupal.org/node/516354

zerolab:
Nice solution! Added a couple of very minor comments.

zerolab:
LGTM!

Issue #287, opened 2016-11-18
Backend editing on multiple Instances (local/stage/prod)


pheadeaux:
As admin, when working on an other environment than production, "editing content" will always send me to the production domain.

I'm using a acquia hosting setup. There is a dev environment, stage environment, prod environment, and a local development environment (each configered with different domains). When logged in on `http://stage.domain.com/admin/content/` every link to `/node/id/edit` will be replaced with `https://prod.domain.com/node/id/edit` when a domain source is set to this very node.

`http://stage.domain.com` is a proper domain alias to the default domain.

To actually edit the node I have to copy the link and rewrite the `prod` to `stage` in this case.

I followed the link creation to `lib/Drupal/Core/Routing/UrlGenerator.php:320` `$options['base_url']`. But there is noting wrong there, I guess.

So is this a misconfiguration in my setup? or a intended design?

I searched the github issues, the drupal issues and (old) install instructions.

Great module, thanks for the effort!

agentrickard:
This is probably a missing piece in Domain Source -- which is still alpha and hasn't really been given much testing.

It means that DomainSourcePathProcessor::processOutbound() needs to be sensitive to the alias status.

agentrickard:
On review, this is a bit of a complex problem, because in this case, you probably want the entire set of domain references changes to match the alias pattern of the test environment.

Here's how we handled that in Drupal 7:

https://www.drupal.org/node/1096988 (See the section on Managing Site Development.)

I had an approach that worked in D7 worked into code as well: https://www.drupal.org/node/1303616

I think this is something that needs to be handled in a custom module at the moment, but looking through the core code, I don't know that there is an entity_load_alter() anymore.

There is, however, hook_domain_source_alter() which you can leverage in a custom module to enforce the behavior you want. However, because the results of that function call Domain::getPath(), I don't think you can alter that result directly.

I think in an ideal world, we might ass a config setting in settings.php that indicates the Alias set to use, and then we can store the set as part of the Alias.

I am open to ideas, but here's a function to get you started.

```
function domain_alias_domain_source_alter($source, $path, $options) {
  $active = \Drupal::service('domain.negotiator')->getActiveDomain();
  if ($active->id() == $source->id() && $active->getMatchType() == DomainNegotiator::DOMAIN_MATCH_ALIAS) {
    $hostname = $active->get('alias');
    if ($hostname && !$active->getRedirect()) {
      $source->setHostname($hostname);
      $source->setPath();
    }
  }
}
[...]

PR #344, opened 2017-04-05
287 alias

agentrickard:
Fix for #287.

See the README for details.

Still TODO:

* Update path to add new field to schema
* Add config schema
* Tests
* UI improvements

### DIFF
--- a/domain_alias/domain_alias.module
+++ b/domain_alias/domain_alias.module
@@ -84,7 +84,7 @@ function domain_alias_domain_load($entities) {
         if ($environment_alias = $loader->loadByEnvironmentMatch($domain, $active->alias->getEnvironment())) {
           // @TODO: handle wildcard patterns.
           $pattern = $environment_alias->getPattern();
-          if (substr_count($pattern, '*' < 1)) {
+          if (substr_count($pattern, '*') < 1) {
             $domain->setHostname($pattern);
             $domain->setPath();
             $domain->setUrl();
--- a/domain_alias/domain_alias.module
+++ b/domain_alias/domain_alias.module
@@ -72,6 +72,7 @@ function domain_alias_domain_operations(DomainInterface $domain, AccountInterfac
  */
 function domain_alias_domain_load($entities) {
   static $enabled;
+  $enabled = TRUE;
   // We can only perform meaningful actions if url.site is a cache context. Otherwise,
   // the render process ignores our changes.
   if (!isset($enabled)) {
--- a/domain_alias/domain_alias.module
+++ b/domain_alias/domain_alias.module
@@ -110,11 +110,29 @@ function domain_alias_domain_load($entities) {
           }
           else {
             // Do a wildcard replacement based on the current request.
-            // @TODO: This doesn't handle ports right now.
             $request = $negotiator->negotiateActiveHostname();
+            // First, check for a wildcard port.
+            if (substr_count($pattern, ':*') > 0) {
+              // Do not replace ports unless they are nonstandard. See
+              // \Symfony\Component\HttpFoundation\Request\getHttpHost().
+              if (substr_count($request, ':') > 0) {
+                $search = explode(':', $pattern);
+                $replace = explode(':', $request);
+                if (!empty($search[1]) && !empty($replace[1])) {
+                  $pattern = str_replace(':' . $search[1], ':' . $replace[1], $pattern);
+                }
+              }
+              // If no port wildcard, then remove the port entirely.
+              else {
+                $pattern = str_replace(':*', '', $pattern);
+              }
+            }
             $search = explode('.', $pattern);
             $replace = explode('.', $request);
-            $pattern = str_replace($search, $replace, $pattern);
+            // This routine only works if the hostnames match length.
+            if (count($search) == count($replace)) {
+              $pattern = str_replace($search, $replace, $pattern);
+            }
             // Do not let the domain loop back on itself.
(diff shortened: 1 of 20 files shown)


domain_alias/domain_alias.module
```diff
@@ -64,3 +66,45 @@ function domain_alias_domain_operations(DomainInterface $domain, AccountInterfac
...
+/**
+ * Implements hook_ENTITY_TYPE_load().
+ */
+function domain_alias_domain_load($entities) {
+  static $enabled;
+  // We can only perform meaningful actions if url.site is a cache context. Otherwise,
+  // the render process ignores our changes.
+  $required_cache_contexts = \Drupal::getContainer()->getParameter('renderer.config')['required_cache_contexts'];
+  if ($enabled === FALSE || (!in_array('url.site', $required_cache_contexts) && !in_array('url', $required_cache_contexts))) {
+    $enabled = FALSE;
```
zerolab: This looks unneeded

agentrickard: I dropped this in here for two reasons:

* To prevent running the code when the site hasn't been configured to leverage it. That's a performance gain, and the code should be further optimized for that, since we load the container multiple times. We should wrap this in an isset() check for $enabled.
* To make the config change a hard dependency, which makes support easier.

Perhaps I'm overthinking it, though.

zerolab: Ah, I see now. It does make sense.

agentrickard:
I just pushed a change that makes the code more clear in the case of the cache context.

agentrickard:
@zerolab can you see if this new test works for you locally? Travis seems to be getting hung up on the call to the alias domain (which should be four.example.com).

This is working fine for me locally.

Travis error:

```
Drupal\Tests\domain_alias\Functional\DomainAliasEnvironmentTest::testDomainAliasEnvironments
    GuzzleHttp\Exception\ConnectException: cURL error 7: couldn't connect to
    host (see http://curl.haxx.se/libcurl/c/libcurl-errors.html)
```

Which tracks back to line 90 of DomainAliasEnvironmentTest.

zerolab:
Locally I got:

```
Drupal\Tests\domain_alias\Functional\DomainAliasEnvironmentTest::testDomainAliasEnvironments
    GuzzleHttp\Exception\ConnectException: cURL error 7: Failed to connect to
    four.example.com port 80: Connection refused
```

The issue here is that we are running the site on port 8080, but it tries to connect on 80. Interestingly enough the Travis error contains neither the domain nor the port.

Will try to look into it.

agentrickard:
I see. We should be able to finesse that in the test.

zerolab:
Much better!

Still failing on 8.3.x with:

```
Drupal\Tests\domain_alias\Functional\DomainAliasEnvironmentTest::testDomainAliasEnvironments
    Behat\Mink\Exception\ExpectationException: Link found
```

agentrickard:
Need to apply the same logic to the link url.

agentrickard:
We might want to set the port in line 45 instead, where we save the pattern.

```
    foreach ($domains as $domain) {
      $values = [
        'domain_id' => $domain->id(),
        'pattern' => array_shift($patterns) . $domain->getPort(),
        'redirect' => 0,
        'environment' => 'local',
      ];
      $this->createDomainAlias($values);
    }
```

agentrickard:
I don't really understand why this would pass on 8.2 and fail on 8.3.

agentrickard:
This may be a cache issue, since we rely on url.site as a CacheContext. I can modify the test to account for that.

agentrickard:
I don't think it's caching, since the blocks are all set to cache by URL.

zerolab:
I think the issue is that the for loop on [line 84](https://github.com/agentrickard/domain/pull/344/files#diff-7e281d560ce4a0be4d0d7125ccabbbf6R83) returns:

```
example.com:8080
one.example.com:8080
two.example.com:8080
two.example.com:8080
```

See https://travis-ci.org/zerolab/domain/builds/221500151 -- I added a var_dump:

```php
    // For a non-aliased request, the url list should be normal.
    $this->drupalGet($domain->getPath());
    foreach ($domains as $domain) {
      $this->assertSession()->assertEscaped($domain->getHostname());
      var_dump($domain->getHostname());
      $this->assertSession()->linkByHrefExists($domain->getPath(), 0, 'Link found');
    }
```

agentrickard:
When I run that I get:

```
string(11) "example.com"
string(15) "one.example.com"
string(15) "two.example.com"
```

zerolab:
argh, red herring!
The failing bit is the final for loop - https://travis-ci.org/zerolab/domain/builds/221555768

It checks for four.example.com, however only
```
example.com:8080
one.example.com:8080
two.example.com:8080
```
are present.

agentrickard:
It may be that the rewrite isn't firing because the CacheContext 'url.site' isn't present. It is on my local (I think).

I couldn't find an obvious way to add that service setting during a test run.

zerolab:
I am not sure how to do it from the test itself, but I tried injecting that in `sites/default/services.yml` but it failed in Travis https://travis-ci.org/zerolab/domain/builds/221635093. 

I think this needs to happen through a Travis setup script, rather than in yml (https://github.com/zerolab/domain/commit/1891324e6aaa6f250a661d95ef09b0e66f5d91a0)

Ran out of time. Will revisit tonight

agentrickard:
That's what I was wondering. Let's try that and see what it does.

PR #351, opened 2017-07-30
Issue #2892612 by SpadXIII: Domain source trips over deleted field_do…

agentrickard:
…main_source


agentrickard:
This test is not failing for me locally.

PR #356, opened 2017-08-10
More granular permissions for Domain Alias.

agentrickard:
Here's a start. Needs a test and perhaps better checking of the access logic.


agentrickard:
This behavior is already covered by the tests in testDomainListBuilder().

PR #357, opened 2017-08-11
Cleans up handling of getActiveDomain().

agentrickard:


agentrickard:
These need t stay as-is for now.

PR #358, opened 2017-08-24
Validate hook

agentrickard:
Fixes the issue documented in https://www.drupal.org/node/2884470


agentrickard:
I really don't understand these test fails.

agentrickard:
Schema should be inherited. What the #!@$ changed in core?

zerolab:
Could it be related to https://www.drupal.org/node/2870971 ?

agentrickard:
Yup. That's it. I just couldn't find the change notice because this manifests as a schema error, and searching for 'schema' doesn't pull up this record.

Issue #337, opened 2017-02-16
make checking in domain_access_entity_field_access() more strict 


andypost:
Currently access checking is separate for UserInterface but else should check for NodeInterface
Otherwise any same named field got this access checking
But this permissions (publish to any domain) are related to nodes only

agentrickard:
We seem to handle this fine now.

```
/**
 * Implements hook_entity_field_access().
 */
function domain_access_entity_field_access($operation, FieldDefinitionInterface $field_definition, AccountInterface $account, FieldItemListInterface $items = NULL) {
  // Hide the domain access fields from the entity add/edit forms
  // when the user cannot access them.
  if ($operation != 'edit') {
    return AccessResult::neutral();
  }

  // The entity the field is attached to.
  $entity = $items->getEntity();

  if ($field_definition->getName() == DOMAIN_ACCESS_FIELD) {
    if ($entity instanceof User) {
      $access = AccessResult::allowedIfHasPermissions($account, [
        'assign domain editors',
        'assign editors to any domain',
      ], 'OR');
    }
    else {
      // Treat any other entity as content.
      $access = AccessResult::allowedIfHasPermissions($account, [
        'publish to any domain',
        'publish to any assigned domain',
      ], 'OR');
    }

    // allowedIfHasPermissions returns allowed() or neutral().
    // In this case, we want it to be forbidden,
    // if user doesn't have the permissions above.
    if (!$access->isAllowed()) {
      return AccessResult::forbidden();
    }
  }
  // Check permissions on the All Affiliates field.
  elseif ($field_definition->getName() == DOMAIN_ACCESS_ALL_FIELD) {
    if ($entity instanceof User) {
      return AccessResult::forbiddenIf(!$account->hasPermission('assign editors to any domain'));
    }

    // Treat any other entity as content.
    return AccessResult::forbiddenIf(!$account->hasPermission('publish to any domain'));
  }

  return AccessResult::neutral();
}
```

andypost:
here's a 2 troubles
- `$entity instanceof User` should check for `UserInterface`
-  "// Treat any other entity as content." this is wrong because domain access fields attached to nodes only and could break contrib like https://github.com/skilld-labs/domain_entity

Later is not issue cos we using other names for fields

agentrickard:
Thanks!

PR #360, opened 2017-08-31
Fix for field access checking.

agentrickard:
See #337

### DIFF
--- a/domain_access/domain_access.module
+++ b/domain_access/domain_access.module
@@ -11,7 +11,6 @@ use Drupal\node\NodeInterface;
 use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Session\AccountInterface;
-use Drupal\user\Entity\User;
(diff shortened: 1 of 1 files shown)


agentrickard:
@andypost This should fix the issue. I used AccountInterface because there may be cases where that is implemented, and UserInterface inherits from it.

andypost:
Looks great to me! Travis needs kinda `echo "sendmail_path=/bin/true" >> /etc/php7/conf.d/xx-mail.ini`

zerolab:
Thanks for the suggestion @andypost, that is kind there, in a slightly different form -- https://github.com/agentrickard/domain/blob/8.x-1.x/.travis.yml#L72

andypost:
@zerolab thanx, then it very strange that tests fail on site install

agentrickard:
That might just be a Travis bug. It passed in all other contexts. Retesting.

agentrickard:
Test passed on retry and all tests pass locally.

PR #361, opened 2017-09-01
Fix for issues reported in https://www.drupal.org/node/2880112

agentrickard:
This will need some review and tests. It seems to be a possible solution for https://www.drupal.org/node/2880112 and https://www.drupal.org/node/2894904


domain/src/DomainRedirectResponse.php
```diff
@@ -0,0 +1,113 @@
...
+    $base_parts = parse_url($base_url);
+
+    if (empty($url_parts['host'])) {
+      throw new \InvalidArgumentException('A path was passed when a fully qualified domain was expected.');
+    }
+
+    // Check that the requested $url is registered.
+    $negotiator = \Drupal::service('domain.negotiator');
+    $registered_domain = $negotiator->isRegisteredDomain($url_parts['host']);
+
```
agentrickard, line 124: Add a check here against Trusted Host Patterns, if set. 

Update documentation to encourage proper use of those settings,

agentrickard:
I HAVE NO $!@%$!@^%$!@ IDEA WHY THIS TEST FAILS ON TRAVIS.

PR #366, opened 2017-09-21
Attempts to fix the middleware issue reported by @michielnugter.

agentrickard:


domain_config/src/DomainConfigOverrider.php
```diff
@@ -165,18 +172,20 @@ public function getCacheableMetadata($name) {
   protected function initiateContext() {
     // Prevent infinite lookups by caching the request. Since the _construct()
     // is called for each lookup, this is more efficient.
-    static $context;
-    if ($context) {
-      return;
-    }
-    $context++;
+    $this->contextSet = TRUE;
```
mnugter, line 177: The patch splits this check into 2 methods. I think it's better to keep everything here as getCacheableMetadata() also calls initiateContext.

If a check outside of the method is required a change to a class static and self::$context could be a better way to go.

mnugter (commented):
The patch fixes the problem so it's a good start!

I have added 1 note on the contextSet though

domain_config/src/DomainConfigOverrider.php
agentrickard: I can't recall what in the call stack made it possible that Cache would be initiated without running the overrides. Possibly registration of the event handler.

Are there use-cases for calling this outside of it's current scope? Perhaps for modules trying to deal with the hook issue?

mnugter:
I  created a test that passes with the fix and failes without.

I had to convert to a Functional test to achieve it. I attached a patch with the diff from this branch. I don't know a better way as I cannot push changes here.

the page_attachments_alter hook does always get called so I had to keep the original hook and check on the changed property there. Let me know what you think.

[2896434-interdiff.patch.zip](https://github.com/agentrickard/domain/files/1358821/2896434-interdiff.patch.zip)

agentrickard:
That looks really good. Nice work!

agentrickard:
I had to remove one assertion from the test, which I couldn't get to work. I think this proves that our shim (the load of alter hooks directly in the Negotiator) works.

mnugter:
Nice! The removed test is actually covered as well (slightly indirect) right now so no coverage is lost here I think.

Anything more todo before this can be merged?

mnugter (approved):

agentrickard:
Nope. Let's commit this and work on the fallout later.

PR #370, opened 2017-10-18
Fix error when getting user domains for anonymous user

piggito:
I get to this error when installing a new site using config_installer. During this process node access are checked for anonymous user which doesn't have the field_domain_access so it was triggering an InvalidArgumentException.
Then I'm adding a validation to check the entity to have the field before getting it.

### DIFF
--- a/domain_access/src/DomainAccessManager.php
+++ b/domain_access/src/DomainAccessManager.php
@@ -52,7 +52,7 @@ class DomainAccessManager implements DomainAccessManagerInterface {
       return $list;
     }
     // Get the values of an entity.
-    $values = $entity->get($field_name);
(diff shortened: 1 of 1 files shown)


agentrickard:
Interesting. I suspect there are other cases where we have to do the same thing. 

Can you also check Domain Source for this type of thing?

agentrickard:
See especially the note at the top of DomainAccessManager:

```
@TODO: It is possible that this class may become a subclass of the
 * DomainElementManager, however, the use-case is separate as far as I can tell.
```

At the least, we have to fix this in both places. Might be better to make DomainAccessManager a subclass of DomainElementManager.

agentrickard:
Looks like those two classes shouldn't be merged, so let's just be sure to cover both cases.

Issue #378, opened 2017-11-03
Removing a domain does not remove its aliases


zerolab:
This makes for an interesting disconnect and validation error if you remove a domain, add it again with a different machine name, then try to create one the previous aliases.

I think this needs to implement `hook_ENTITY_TYPE_delete()` for domain and remove all attached aliases.

agentrickard:
Agreed.

agentrickard:
I wonder how we deal with this for Domain Config?

agentrickard:
Thanks to jet lag, I was up early and fixed this, with a test. Could use a review.

PR #381, opened 2017-11-08
Delete aliases on domain deletion.

agentrickard:
Fix for [#378]

### DIFF
--- a/domain_alias/tests/src/Kernel/DomainAliasDomainDeleteTest.php
+++ b/domain_alias/tests/src/Kernel/DomainAliasDomainDeleteTest.php
@@ -1,0 +1,83 @@
+<?php
+
+namespace Drupal\Tests\domain_alias\Kernel;
+
+use Drupal\Component\Render\FormattableMarkup;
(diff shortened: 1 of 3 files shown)


zerolab:
LGTM, thank you!

domain_alias/tests/src/Kernel/DomainAliasDomainDeleteTest.php
```diff
@@ -0,0 +1,83 @@
...
+class DomainAliasDomainDeleteTest extends DomainTestBase {
+
+  use DomainAliasTestTrait;
+
+  /**
+   * Modules to enable.
+   *
+   * @var array
+   */
+  public static $modules = array('domain', 'domain_alias');
```
zerolab, line 23: nitpicking: array shorthand?
Will need to submit a PR that runs phpcs with the Drupal 8 standards

agentrickard: Yeah, that's a copy paste from another test that we can clean up later.

PR #384, opened 2017-11-27
Issue #2925629 by Alan-H: Saving after sort order change overwrites h…

agentrickard:
Patch from https://www.drupal.org/project/domain/issues/2925629


agentrickard:
I prefer the approach in #385

PR #393, opened 2017-12-09
Domain content tests

agentrickard:
There is a lot going on here. The last missing piece, I think, is the failing test.

For some reason, the Access handler assigned to the Domain Content Views does not hide the page callback, just the View content. 

That seems wrong to me. Why wouldn't a Views access handler deny page access?

Perhaps the page needs to be added to the menu?


agentrickard:
No. You can't add these to a menu:

``` Views cannot create normal menu items for paths with a % in them. ```

Is this a bug in Views? Is it acceptable?

@zerolab any ideas?

**Comment by zerolab** [comment]
Hey @agentrickard,

The error it is thrown by `Drupal\views\Plugin\views\display\Page::validateOptionsForm()`. The behaviour has been around since Views 2 / D6.

Looking at the commits, did you fix the access issue in 7b879d4 ?

**Comment by agentrickard** [comment]
Well, the access error is fixed -- sort of -- in that when you go to a View page for a domain you cannot edit, the view is hidden but the page title is still displayed instead of a Drupal 403 error.

This test fails because it expects 403 and instaed gets back a 200 with an empty page.

So why isn't Views enforcing the access rule on the route?

**Comment by agentrickard** [comment]
I forgot to add domain content to the test runner.

**Comment by agentrickard** [comment]
Thanks. That makes a lot of sense. I figured Views would do that for me.

So this is an interesting case. We have both a permission (from Domain Content) that allows users to view these pages. And we have this "can edit" access from Domain Access.

The Domain Access handler here is meant to be fairly generic -- that is, it should work outside the context of Domain Content. What I'm torn about is whether we need to enforce the permission as well within the View.

See the logic in `DomainContentController::contentList()`. That page is restricted by the permission `access domain content`.

So I'm trying to duplicate that access logic on the View. Perhaps that should go in Domain Content and not in Domain Access? Or I could make two sets of access plugins, one which combines the permission and one which doesn't.

Ideas? Examples?

**Comment by zerolab** [comment]
OTOH, the simplest is to have both. The Domain Access one is generic and pretty much says "I belong to this domain", while the Domain Content views access check should check the "access domain permission" as well as belonging to the current domain.

In an ideal world, we'd have a Domain permission access plugin that would discover all the permissions declared by the Domain modules and then be a proxy for `PermissionAccessCheck` for the current domain.

Currently deep in a content audit for a migration. Will have a think on the way back home and report back.

**Comment by agentrickard** [comment]
Thanks. I'm having good success with the service approach. What I can do is make extra Access handlers inside Domain Content that adds the extra permission check.  That should work fine.

**Comment by agentrickard** [comment]
The code I just posted is about 70% done. We need to pull the path argument from the route.

**Comment by agentrickard** [comment]
I think I'm hitting a cache issue in the test fail, which I can solve.

But I wonder, is this the right approach? Views doesn't really seem to want to pass arguments to the Route handler -- the $arg_0 bit seems very hacky,

@zerolab 

Would this be better off as an argument (contextual filter) handler that displays "Access Denied" when the filter fails to validate?

**Comment by zerolab** [comment]
You should be able to inject the current request service, and from there it should be possible to get the `_route_params`.

Will see if I can make some headway tonight. (sorry for the silence the other day :( )

**Comment by agentrickard** [comment]
It didn't occur to me to inject that service, but that can be solved with passign params to the route / access handler -- which I do now.

The issue that makes me nervous is that we don't _know for certain_ what position the argument is in.

File: domain_access/src/Access/DomainAccessViewsAccess.php [hunk 1]
```diff
@@ -0,0 +1,120 @@
...
+    $this->entityTypeManager = $entity_type_manager;
+    $this->domainStorage = $this->entityTypeManager->getStorage('domain');
+    $this->userStorage = $this->entityTypeManager->getStorage('user');
+    $this->manager = $manager;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function access(Route $route, AccountInterface $account, $arg_0 = 'all_affiliates') {
```
agentrickard: Here's the $arg_0, which is passed from the view.

agentrickard:
I figured out the test fails. Pushing for re-test.

agentrickard:
I have no idea why the views are failing to install properly.

agentrickard:
Because you can't include a dependency on the module that provides the View. **sigh**

agentrickard:
OK, this is working. I just want to expand the Permission test, and then I think it's good-to-go.

PR #394, opened 2017-12-18
Response check

agentrickard:
Validates URL responses when registering domains. Can be disabled.


agentrickard:
These tests are passing locally.

PR #400, opened 2018-01-22
Add missing domain_source.views.schema.yml for domain_source filter

SpadXIII:
The domain_source.views.schema.yml file is required by views, otherwise it does not seem to know what to do with a field_domain_source_target_id filter.


agentrickard:
How does this compare to #396?

agentrickard:
Duplicate.

PR #401, opened 2018-02-01
Issue #2941319 by agentrickard, hbensalem, yo30: Error: Call to undef…

agentrickard:
See https://www.drupal.org/project/domain/issues/2941319#comment-12462368


agentrickard:
This was fixed in core.

PR #406, opened 2018-04-03
Force is_default to boolean value.

dpagini:
Pull Request for http://drupal.org/node/2957997

Sorry, I hope I'm following an appropriate workflow here... posting on d.o and then the PR on Github.

### DIFF
--- a/domain/src/Entity/Domain.php
+++ b/domain/src/Entity/Domain.php
@@ -100,7 +100,7 @@ class Domain extends ConfigEntityBase implements DomainInterface {
    *
    * @var bool
    */
-  protected $is_default;
+  protected $is_default = FALSE;
 
   /**
    * The domain record protocol (e.g. http://).
@@ -211,12 +211,12 @@ class Domain extends ConfigEntityBase implements DomainInterface {
       // Swap the current default.
       /** @var self $default */
       if ($default = \Drupal::service('entity_type.manager')->getStorage('domain')->loadDefaultDomain()) {
-        $default->is_default = 0;
+        $default->is_default = FALSE;
         $default->setHostname($default->getCanonical());
         $default->save();
       }
       // Save the new default.
-      $this->is_default = 1;
+      $this->is_default = TRUE;
       $this->setHostname($this->getCanonical());
       $this->save();
     }
@@ -332,11 +332,11 @@ class Domain extends ConfigEntityBase implements DomainInterface {
     /** @var self $default */
     $default = $storage->loadDefaultDomain();
     if (!$default) {
-      $this->is_default = 1;
+      $this->is_default = TRUE;
     }
     elseif ($this->is_default && $default->getDomainId() != $this->getDomainId()) {
       // Swap the current default.
-      $default->is_default = 0;
+      $default->is_default = FALSE;
       $default->save();
     }
     // Ensures we have a proper domain_id.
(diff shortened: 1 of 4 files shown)


dpagini:
At first I was unsure of how to demonstrate this with a test. It _may_ make sense to use a dummy `domain.record.____.yml` record into the test module, and see if that will pass test before/after this PR...?
I'd love to get some feedback on this approach first just to see if there was some reason it was done this way and this new approach won't work for some reason. 
Just FYI, I took some inspiration from the Menu.php entity in core which exports a boolean property, and was how I went with this proposed fix.

agentrickard:
This is probably a remnant from some very early code work before I understood what the schema API was really doing. Explicit TRUE/FALSE seems right here, if that's what the schema is expecting.

I'd love to see a test that failed due to the lack of casting, though.

agentrickard:
(Hit the wrong button)

dpagini:
So, I'm adding a phony domain record to the `domain_test` module. This is _maybe sort of_ (?) cheating, b/c I'm expecting this to now start failing the `DomainNegotiatorTest` class, which installs this module in the `::setUp()` method.

dpagini:
Nice! It failed... I'm going to push up my commit fix on top of this one now, which should show a passing test. Will this be a satisfactory test scenario for you?

agentrickard:
Nice.

dpagini:
Hmm, sorry, I think I messed up my commits... anyways, going to keep working through, sorry, haha.

dpagini:
Ok, so I noticed many of the tests actually check first that there are no domains created in the system, so by adding one to the test module, that instantly fails that assertion. I am creating a new module called `domain_config_test` and providing that dummy config record, and then loading those records in a KernelTest, which I am expecting to throw the error. In my `DomainConfigTest` I am also going to load the `views` module and the `domain_test` which should hopefully show the scenario in #407 as well. Just going to wait on the tests to run now.

agentrickard:
That makes sense. Making a new module should not trigger any test errors if you avoid using the domainTableIsEmpty() test in DomainTestTrait. 

The new module shouldn't be enabled when the other tests run.

dpagini:
Ok, I am expecting 8cf3279 to fail via:
https://travis-ci.org/agentrickard/domain/builds/362151870

So [here is the fail](https://travis-ci.org/agentrickard/domain/jobs/362151872) I'm looking for:
screen shot 2018-04-04 at 10 21 12 am

_Seems there's a few hiccups in the testing... some timeout, some strange unrelated test fails, etc._

And the following commit d03d64f to pass via:
https://travis-ci.org/agentrickard/domain/builds/362187596

I gave up on testing 407 in here, but if this goes in, I think I can use the same test module to load a views config to prove out that fix...

dpagini:
hey @agentrickard - this run is looking good to me. There is a failure, but [this one](https://travis-ci.org/agentrickard/domain/jobs/362187600#L822) seems a little bit more like a hiccup to me than a red flag. How would you like me to proceed here?

screen shot 2018-04-04 at 12 01 18 pm

agentrickard:
That one fails occasionally on Travis. If it only failed for one run, it can be ignored. 

I'll rerun all the tests locally anyway.

dpagini:
Great, sounds good. Sorry for all the Travis tests I ended up running for this, haha. I'll keep an eye on this guy and let me know if you need me to address anything else here, thanks @agentrickard!

agentrickard:
You are writing actual tests, so you get all the credit.  And Travis can handle it.

dpagini:
Thanks @agentrickard!

PR #407, opened 2018-04-03
Provide domain access config schema for views.

dpagini:
see https://www.drupal.org/project/domain/issues/2957999

### DIFF
--- a/domain/config/schema/domain.schema.yml
+++ b/domain/config/schema/domain.schema.yml
@@ -83,3 +83,13 @@ condition.plugin.domain:
       sequence:
         type: string
 
+views.access.domain:
+  type: mapping
+  label: 'Domains'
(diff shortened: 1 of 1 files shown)


agentrickard:
IS this all that we're missing right now? I thought there was more?

dpagini:
So I have a view that uses this access mechanism that is failing when I'm trying to run tests. I'm honestly not sure if there would be more to provide schemas for...

agentrickard:
Use the config_inspector module.

dpagini:
Yeah, I mean it's the only config that I'm using that's giving me a problem. I don't use the whole suite of modules, mostly just `domain` and `domain_access` I believe. I wouldn't have even seen the views issue unless I had a view that used domain to limit access, you know?

PR #408, opened 2018-04-18
Issue #2961719 by h3rj4n: Library Discovery Collector cache per domain

agentrickard:


agentrickard:
Note that PHP 7 tests will fail for Drupal 8.6

PR #411, opened 2018-04-18
Issue #2940296 by david.gil, Berdir: Combination of language negotiat…

agentrickard:
…ion and path aliasing can cause a corrupted route cache, 404s


agentrickard:
The intermittent test fail is odd. I'm going to merge this and then work on https://www.drupal.org/project/domain/issues/2976127

PR #412, opened 2018-04-18
Issue #2959309 by davidgrayston: Call to undefined function base_path()

agentrickard:


agentrickard:
This failure happens occasionally on Travis.

PR #413, opened 2018-04-19
Delete the Domain Alpha module. Requires a cache rebuild.

agentrickard:
See https://www.drupal.org/project/domain/issues/2960228


rivimey:
@agentrickard perhaps this would be a good time to action this PR?

agentrickard:
Yes, before the beta, at the least. My concern is that it requires people to uninstall the module before it can be deleted.

zerolab:
I think that should be an upgrade consideration in the release notes.

Unless you want to add a `hook_update_n` implementation in one release, then actually remove the module in a followup one..

```php
/**
 * Uninstall Domain Alpha, if installed.
 */
function domain_update_8002() {
  \Drupal::service('module_installer')->uninstall(['domain_alpha']);
}
```

agentrickard:
I was just going to add it in the release notes, because even an uninstall like this won't stop people from re-enabling it.

The problem is that few people read release notes.

agentrickard:
That said, the next release is also going to require Drupal 8.5+, so reading the notes is pretty imperative.

zerolab:
> The problem is that few people read release notes.

Heh, true. But you cannot account for everyone. And given the module is in alpha, things can still break. I think adding the release notes will be enough. And for the inevitable handful of people that don't bother to read the upgrade notes, well.. you did your best :)

rivimey:
See also https://www.drupal.org/project/domain/issues/2960228

PR #420, opened 2018-06-01
Issue #2976899 by idebr: [Domain source] Follow Drupal coding standards

ideb:


agentrickard:
These test fails are unrelated to this work.

domain_source/domain_source.module
```diff
@@ -166,7 +169,7 @@ function domain_source_form_validate($element, \Drupal\Core\Form\FormStateInterf
 /**
  * Redirect form submissions to other domains.
  */
-function domain_source_form_submit(&$form, \Drupal\Core\Form\FormStateInterface $form_state) {
+function domain_source_form_submit(&$form, FormStateInterface $form_state) {
```
agentrickard, line 172: Do we have to type hint &$form?

domain_source/domain_source.module
```diff
@@ -138,7 +141,7 @@ function domain_source_form_alter(&$form, &$form_state, $form_id) {
 /**
  * Validate form submissions.
  */
-function domain_source_form_validate($element, \Drupal\Core\Form\FormStateInterface $form_state) {
+function domain_source_form_validate($element, FormStateInterface $form_state) {
```
agentrickard, line 144: Do we have to type hint $element?

agentrickard (commented):
Two simple questions about type hints

domain_source/domain_source.module
ideb: FormSubmitterInterface::doSubmitForm does not typehint the &$form array, but it is assumed to be an array so I added a type cast to make this explicit

domain_source/domain_source.module
ideb: FormValidatorInterface::executeValidateHandlers does not typehint the &$form array, but it is assumed to be an array so I added a type cast to make this explicit

agentrickard (changes requested):
I would remove these changes. I looked at core, and they don't type hint these vars.

```
/**
 * Implements hook_form_BASE_FORM_ID_alter() for \Drupal\node\NodeForm.
 */
function path_form_node_form_alter(&$form, FormStateInterface $form_state) {
```

agentrickard:
Thanks! I'll wait for the tests here and then clean up anything that breaks.

PR #422, opened 2018-06-04
Note exceptions thrown; improve doc of options_select

rivimey:
Phpstorm noted that that `domain_source_confirm_fields()` threw undocumented exceptions.

There was also a lack of clarity around the code "Tell the form system how to behave. Default to radio buttons." which I have restructured and added a type hint to so that `setComponent()` is findable.

I also wondered whether `\Drupal::entityTypeManager()` can be changed to something like: `\Drupal::service('entity_type.manager')` as done in `domain_source_get()` lower down. 

Let me know if this sort of change is not appropriate ... in particular not sure what the policy around documenting exceptions is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,7 +118,6 @@ marked with [x] are considered complete.
 - [ ] Caching strategies in DomainNegotiator
 - [ ] Caching strategies in DomainConfigOverrides
 - [ ] Cache in the DomainAccessManager
-- [ ] Add filter options to domain_access and domain_source views
 - [ ] Proper handling of default node values
 - [ ] Do not allow actions to be edited?
 - [o] Recreate the Domain Theme module -- see https://www.drupal.org/project/domain_theme_switch
--- a/domain_config/README.md
+++ b/domain_config/README.md
@@ -45,9 +45,9 @@ langcode: en
 default_langcode: en
 ```
 
-An override file should contain only the specific data that you want to override.
-To override the name, the only line needed is the one beginning ```name:```.
-The resulting override looks like this:
+An override file should contain only the specific data that you want to
+override. To override the name, the only line needed is the one beginning
+ ```name:```. The resulting override looks like this:
 
 ```YAML
 name: Three
@@ -84,13 +84,23 @@ settings.php overrides
 ----------------------
 
 For environment-specific or sensitive overrides, use the settings.php method.
-In the above case, add
-`$config['domain.config.three_example_com.en.system.site']['name'] = "My special site";`
-to your local settings.php file. This will ensure that the `three.example.com`
-domain gets the correct value regardless of other module overrides.
+In the above case, add:
 
-Read more about the "Configuration override system"
-at https://www.drupal.org/node/1928898.
+`$config['domain.config.three_example_com.en.system.site']['name'] = "My special site";`
+
+to your local settings.php file. This will ensure that the `three.example.com`
+domain gets the correct value regardless of other module overrides. If you do
+not have a language set the config system will fallback on the domain-specific
+setting (note the missing '.en'):
+
+`$config['domain.config.three_example_com.system.site']['name'] = "My special site";`
+
+To set nested values you need to nest the config array:
+
+`$config['domain.config.three_example_com.system.site']['page']['front'] = '/node/1;`
+
+Read more about the "Configuration override system" at
+https://www.drupal.org/node/1928898.
 
 Installation
 ============
--- a/domain_source/domain_source.module
+++ b/domain_source/domain_source.module
@@ -23,6 +23,9 @@ const DOMAIN_SOURCE_FIELD = 'field_domain_source';
  * @param string $bundle
  *   The bundle being created.
  *
+ * @throws \Drupal\Core\Entity\EntityStorageException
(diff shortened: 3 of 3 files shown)


agentrickard:
Your tests are failing because your HEAD is not up-to-date with the GItHub 8.x-1.x.

Please update / merge your fork so that tests will pass.

domain_source/domain_source.module
```diff
@@ -23,6 +23,9 @@ const DOMAIN_SOURCE_FIELD = 'field_domain_source';
  * @param string $bundle
  *   The bundle being created.
  *
+ * @throws \Drupal\Core\Entity\EntityStorageException
```
agentrickard, line 26: We aren't explicitly throwing these exceptions, so I don't know that we would document them here. Are there cases where Exceptions are thrown? Under what conditions?

domain_source/domain_source.module
```diff
@@ -54,7 +57,11 @@ function domain_source_confirm_fields($entity_type, $bundle) {
   }
 
   // Tell the form system how to behave. Default to radio buttons.
-  if ($display = \Drupal::entityTypeManager()->getStorage('entity_form_display')->load($entity_type . '.' .  $bundle . '.default')) {
+  /** @var $display \Drupal\Core\Entity\Display\EntityDisplayInterface */
+  $display = \Drupal::entityTypeManager()
```
agentrickard, line 61: Use of `\Drupal::entityTypeManager()` is fine, since that's available in `Drupal.php` we should strive for consistency.

agentrickard (commented):
In general, I'd like to tackle the bigger functionality issues before diving into code style. Trying to fix bigger functional issues is compliocat3ed by trying to track these minor changes.

domain_source/domain_source.module
rivimey: The execeptions are thrown by functions called by this function, specifically by entityTypeManager's getStorage() and setComponent(), (IIRC not on work system at present).

My question about 'policy around documenting exceptions' relates to exactly this.

 - if the exceptions are documented, then docblock-aware editors can see that it is possible to receive those exceptions from a call even though there is no explicit throw;
 - if the exceptions are caught and dealt with, then we're good;
 - if neither, a caller can be surprised by a thrown exception.

Happy to follow your lead.

domain_source/domain_source.module
rivimey: ?? confused. The code currently uses both `\Drupal::entityTypeManager()` and `\Drupal::service('entity_type.manager')` so is not consistent now.

rivimey:
My branches were a couple of commits behind, and I have now rebased them and pushed to match current HEAD. Sorry for any inconvenience.

Answering your note about bigger areas: I wanted to start with simple things to get a feel for the module and how it's built. I absolutely want to move on to larger areas, and the items in this PR were only picked on because they were in front of me, so to speak.

agentrickard:
I can work around what helps you get oriented. Glad to have the help.

domain_source/domain_source.module
agentrickard: That's because it was written at two different times, and some of it likely before the EntityTypeManager was a service, so it was updated differently. 

We should standardize on `\Drupal::entityTypeManager()` but that can be in a later cleanup.

domain_source/domain_source.module
agentrickard: My opinion is that any of our code that can throw an exception should be wrapped in TRY/CATCH and then documented. I don't expect this code to throw exceptions, though.

We can address that in a followup.

domain_source/domain_source.module
rivimey: Ok, that's fine.

domain_source/domain_source.module
rivimey: So would it be better to remove the @throws for the moment and deal with the issue as a whole at a later time?

agentrickard: Either that or determine where the errors can be thrown, though we can't set TRY/CATCH if we aren't throwing our own error, can we?

CHANGELOG.md
```diff
@@ -118,7 +118,6 @@ marked with [x] are considered complete.
 - [ ] Caching strategies in DomainNegotiator
 - [ ] Caching strategies in DomainConfigOverrides
 - [ ] Cache in the DomainAccessManager
-- [ ] Add filter options to domain_access and domain_source views
```
agentrickard, line 121: This was already committed to the GitHub version, which hasn't been pushed to d.o. yet because other changes drop support for Drupal 8.4.

domain_config/README.md
```diff
@@ -45,9 +45,9 @@ langcode: en
 default_langcode: en
 ```
 
-An override file should contain only the specific data that you want to override.
-To override the name, the only line needed is the one beginning ```name:```.
-The resulting override looks like this:
+An override file should contain only the specific data that you want to
+override. To override the name, the only line needed is the one beginning
+ ```name:```. The resulting override looks like this:
```
agentrickard, line 50: This has also been committed.

agentrickard:
There's a lot going on in this PR that's just about getting our development habits in sync. I would discard it and make follow-up issues:

* Use EntityTypeManager consistently
* Document exceptions thrown by invoked code

PR #426, opened 2018-06-06
Testing Ruth/8.x 1.x config ui domain switch

rivimey:
For testing


agentrickard:
This should now work with a merge from 8.x-1.x

PR #431, opened 2018-06-15
INFO ONLY: initial drush9 conversion

rivimey:
Ken, thought I'd show you where I'm at with the drush9 conversion of the drush commands.
Most things work except for the delete domain and some forms of verify. Work continuing...

I've changed some of the other interfaces where that seemed reasonable to me. Please let me know if these are not viable. These are in : DomainStorage. DomainValidator

### DIFF
--- a/domain/src/Commands/DomainCommands.php
+++ b/domain/src/Commands/DomainCommands.php
@@ -3,10 +3,12 @@
 namespace Drupal\domain\Commands;
 
 use Consolidation\OutputFormatters\StructuredData\PropertyList;
+use Drupal\Component\Plugin\Exception\PluginException;
 use Drupal\Core\Config\StorageException;
 use Drush\Commands\DrushCommands;
 use Drupal\Component\Utility\Html;
 use Drupal\domain\DomainInterface;
+use Drupal\Component\Serialization\Yaml;
 use Consolidation\OutputFormatters\StructuredData\RowsOfFields;
 
 use Drupal\Core\Entity\EntityStorageException;
@@ -199,6 +201,62 @@ class DomainCommands extends DrushCommands {
   /**
    * Reassign entities of the supplied type to the $policy domain.
    *
+   * @return string[]
+   *   List of entities supporting domain access.
+   */
+  protected function findDomainEnabledEntities() {
+    $entities = [];
+    $entity_manager = \Drupal::entityManager();
+    $field_map = $entity_manager->getFieldMap();
+    foreach($field_map as $type => $fields) {
+      if (array_key_exists(DOMAIN_ACCESS_FIELD, $fields)) {
+        $entities[] = $type;
+      }
+    }
+    return $entities;
+  }
+
+  /**
+   * Reassign entities of the supplied type to the $policy domain.
+   *
+   * @return string[]
+   *   List of entities supporting domain access.
+   */
+  protected function enumerateDomainEntities($entity_type) {
+    $efq = \Drupal::entityQuery($entity_type);
+//    $eq->exists(DOMAIN_ACCESS_FIELD);
+    $result = $efq->execute();
+//    var_dump($efq);
+//    var_dump($result);
+    $entities = [];
+    foreach($result as $item) {
+      $entities = $item;
+    }
+    var_dump($entities);
+    return $entities;
+  }
+
+  /**
+   * Reassign entities of the supplied type to the $policy domain.
+   *
+   * @return string[]
+   *   List of entities supporting domain access.
+   */
+  protected function reassignEntities($entity_type, $new_domain) {
+    $entities = [];
+    $entity_manager = \Drupal::entityManager();
+    $field_map = $entity_manager->getFieldMap();
+    foreach($field_map as $type => $fields) {
+      if (array_key_exists(DOMAIN_ACCESS_FIELD, $fields)) {
+        $entities[] = $type;
+      }
+    }
+    return $entities;
+  }
+
+  /**
+   * Reassign entities of the supplied type to the $policy domain.
+   *
    *
    * @param $entity_type
    *   The name of an entity type, e.g. 'node'.
@@ -212,17 +270,12 @@ class DomainCommands extends DrushCommands {
    *   as single-domain content (FALSE), or just removec (TRUE).
    */
   protected function reassignLinkedEntities($entity_type, $domains, array $options = []) {
-    $entity_manager = \Drupal::entityTypeManager();
-    $entity_definition = $entity_manager->getDefinition($entity_type, FALSE);
-
-    // Domain Access can only use entities that have bundles.
-    if ($entity_definition->getBundleEntityType()) {
-      $list_builder = $entity_manager->getListBuilder($entity_type);
-      $query        = $list_builder->getStorage()->getQuery();
-      $query->exists(DOMAIN_ACCESS_FIELD);
-      $query->sort($this->entityType->getKey('id'));
-
-      $entity_manager;
+    $entity_manager = \Drupal::entityManager();
+    $field_map = $entity_manager->getFieldMap();
+    foreach($field_map as $type => $fields) {
+      if (array_key_exists(DOMAIN_ACCESS_FIELD, $fields)) {
+        // found an entity type that supports domain access.
+      }
     }
   }
 
@@ -317,8 +370,10 @@ class DomainCommands extends DrushCommands {
    * count: All Domains
    * count_active: Active Domains
    * default_id: Default Domain ID
-   * default_host: Default Domain Hostname
-   * scheme: Domain entity fields
+   * default_host: Default Domain hostname
+   * scheme: Fields in Domain entity
+   * domain_access_entities: Domain access entities
+   * all_affiliate_support: All-affiliate support
    * @list-orientation true
    * @format table
    * @throws \Drupal\domain\Commands\DomainCommandException
@@ -365,6 +420,45 @@ class DomainCommands extends DrushCommands {
 
       $rows[$key] = $v;
     }
+
+    // Display which entities are enabled for domain access by checking for the fields.
+    $entity_manager = \Drupal::entityManager();
+    $field_map = $entity_manager->getFieldMap();
+    $domain_entities = [];
+    $domain_affiliate_entities = [];
+    foreach($field_map as $type => $fields) {
+      if (array_key_exists(DOMAIN_ACCESS_FIELD, $fields)) {
+        $domain_entities[] = '+'.$type;
+      }
+      else {
+        try {
+          $def = $entity_manager->getDefinition($type, FALSE);
+        } catch (PluginException $ex) {
+          $domain_entities[] = '{notfound:' . $type . '}';
+          continue;
+        }
+        if ($def->entityClassImplements('Drupal\Core\Entity\FieldableEntityInterface')) {
+          // Entity is fieldable, so could have domain fields.
+          $domain_entities[] = '-' . $type;
+        }
+        elseif ($def->entityClassImplements('Drupal\Core\Entity\ConfigEntityInterface')) {
+          // Entity is configuration entity, which is not fieldable.
+          $domain_entities[] = '!' . $type;
+        }
+        elseif ($def->entityClassImplements('Drupal\Core\Entity\EntityInterface')) {
+          // Entity does not support fields.
+          $domain_entities[] = '/' . $type;
+        }
+      }
+      if (array_key_exists(DOMAIN_ACCESS_ALL_FIELD, $fields)) {
+        $domain_affiliate_entities[] = $type;
+      }
+    }
+    $rows['domain_access_entities'] = implode(', ', $domain_entities);
+    $rows['all_affiliate_support'] = implode(', ', $domain_affiliate_entities);
+
+    $e = $this->enumerateDomainEntities('node');
+    
     return new PropertyList($rows);
   }
 
--- a/domain/src/Commands/DomainCommands.php
+++ b/domain/src/Commands/DomainCommands.php
@@ -238,15 +238,16 @@ class DomainCommands extends DrushCommands {
    * @return string[]
    *   List of entity IDs for the selected domain.
    */
-  protected function enumerateDomainEntities($entity_type, $field, $domain) {
+  protected function enumerateDomainEntities(string $entity_type, string $field, string $domain) {
     $efq = \Drupal::entityQuery($entity_type);
+    // Dont access check or we wont get all of the possible entities moved.
+    $efq->accessCheck(FALSE);
     $efq->condition($field, $domain, '=');
     $result = $efq->execute();
     $entities = [];
     foreach($result as $item) {
-      $entities = $item;
+      $entities[] = $item;
     }
-    //var_dump($entities);
     return $entities;
   }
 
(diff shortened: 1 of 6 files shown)


domain/src/DomainStorage.php
```diff
@@ -82,6 +82,8 @@ public function loadDefaultId() {
 
   /**
    * {@inheritdoc}
+   *
+   * @return \Drupal\domain\DomainInterface|NULL
```
zerolab: these returns should go on the interface

rivimey: Thx for looking!
I've just reviewed this, and these are serving a useful purpose: the interface does have these declarations, but the class {@inheritdoc} is not inheriting the interface but the base class doc, which has EntityInterface.

The net result is that within code manipulating class object refs (e.g. within domain module) these functions return EntityInterface objects, not DomainInterface objects.

I've pushed an update which works for me: let me know if there's a better way.

domain/src/DomainStorage.php
```diff
@@ -98,7 +98,7 @@ public function loadDefaultDomain() {
    * @return \Drupal\domain\DomainInterface[]
    */
   public function loadMultiple(array $ids = NULL) {
-    $domains = $this->loadMultiple($ids);
+    $domains = parent::loadMultiple($ids);
```
agentrickard: If we're calling parent here, do we even need this method override? 

This may be a legacy method from before EntityStorage was standardized.

rivimey: This and some others are my fault: I wanted the functions in DomainStorage class to be declared as returning items of type DomainInterface, so the type checking in phpcs correctly reports on functions and their args.

I am very happy to remove the additional code if there is a better way. Might that way be using @method tags: http://docs.phpdoc.org/references/phpdoc/tags/method.html / https://www.drupal.org/project/coding_standards/issues/2920333 ?

If so the whole loadMultiple function and docblock referenced here could, I think, be reduced to:

    /**
     * @method \Drupal\domain\DomainInterface[] loadMultiple(array $ids = NULL)
     */

agentrickard: The use of parent here is a code smell.

I wouldn't rewrite that, because it confuses developers used to the core API. Returning generic entities are fine. I think you can still type hint on the array elements returned by EntityStorageBase::loadMultiple() if you wanted to.

The `sort` method called by LoadMultipleSorted() does that.

```
  /**
   * {@inheritdoc}
   */
  public function sort(DomainInterface $a, DomainInterface $b) {
    return $a->getWeight() > $b->getWeight();
  }
```

rivimey: I agree about the smell, but the case of sort() here is for parameters, not function results.

Just to be sure, did you realize that I added the whole loadMultiple() function around line 100, not just changed line 101?

I don't really understand "confuses developers used to the core API", given that these functions are being called on domain objects and have the same function signature and results as for the entity api: could you elaborate?

I wonder if it would be useful to meet on drupal slack tomorrow to clarify this and any other issues?

agentrickard: Sure. I can be in Slack. I'm around today, too.

I only noticed you added this method after the initial reply. 

The sort method definition type hints the array elements returned by loadMultiple(), knowing that the base class will return DomainInterface objects as an array.

```
 public function sort(DomainInterface $a, DomainInterface $b)
```

The confusion is that our class has a method that overwrites the base class but merely invokes it as `parent`, which makes me scratch my head as a developer. Why override the base method if it isn't changed?

domain/src/DomainStorage.php
```diff
@@ -94,6 +95,16 @@ public function loadDefaultDomain() {
 
   /**
    * {@inheritdoc}
+   * @return \Drupal\domain\DomainInterface[]
+   */
+  public function loadMultiple(array $ids = NULL) {
```
zerolab: I really think these do not belong here. 
`NodeStorage` does not do this, nor other things in core.
As @agentrickard said, you are merely deferring to the parent and that is an antipattern as far as I kow

rivimey:
No problem, I've pulled all the changes to DomainStorage*

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1108 @@
...
+
+  /**
+   * Get a domain storage object or throw an exception.
+   *
+   * @return \Drupal\domain\DomainStorageInterface
+   *
+   * @throws \Drupal\domain\Commands\DomainCommandException
+   */
+  protected function getStorage() {
+    if ($this->domain_storage) {
```
agentrickard: Probably more consistent to use `!is_null` here.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1108 @@
...
+    }
+    catch (InvalidPluginDefinitionException $e) {
+      throw new DomainCommandException('Unable to get domain: bad storage', $e);
+    }
+
+    return $this->domain_storage;
+  }
+
+  /**
+   * Lookup a string representing a domain config into that config, or throw.
```
agentrickard: THis sentence doesn't read very well. I don't know what it means.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1108 @@
...
+   * 
+   * @param string $argument
```
agentrickard: Does this actually handle both cases? It looks like it only handles machine_name.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1108 @@
...
+    if (!$domain) {
+      throw new DomainCommandException(
+        dt('Domain record could not be found from "!a".', ['!a' => $argument])
+      );
+    }
+    return $domain;
+  }
+
+  /**
+   * Filter a list of domains by excluding
```
agentrickard: By excluding what?

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1108 @@
...
+   *   List of domains.
+   * @param int[] $exclude
+   *   List of domain ID numbers to exclude from the list.
+   *
+   * @return array
+   */
+  protected function filterDomains($domains, $exclude, $initial = []) {
+    $list = $initial;
+    foreach ($domains as $one) {
+      // Not the domain we're deleting, and not the default.
```
agentrickard: Note.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1108 @@
...
+   *
+   * @param \Drupal\domain\DomainInterface $domain
+   *
+   * @return string
+   */
+  protected function checkHTTPResponse(DomainInterface $domain, $validate_url) {
+    // Ensure the url is rebuilt.
+    if ($validate_url) {
+      $this->validateDomain($domain->getHostname());
+    }
```
agentrickard: See `DomainGetResponseTest` for the expected behavior here.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1108 @@
...
+      if (!$domain instanceof DomainInterface) {
+        throw new StorageException('deleting domains: value is not a domain');
+      }
+
+      $hostname = $domain->getHostname();
+      try {
+        $domain->delete();
+      }
+      catch (EntityStorageException $e) {
+        throw new DomainCommandException(dt('Unable to delete domain: @dom',
```
agentrickard: Let's not shorthand `@domain`

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1108 @@
...
+    $validator = \Drupal::service('domain.validator');
+    return $validator->validate($hostname);
+  }
+
+  /**
+   * Deletes a domain record.
+   *
+   * @param DomainInterface[] $domains
+   *   The domain_id to delete. Pass 'all' to delete all records.
+   * @param string[] $delete_options
```
agentrickard, line 186: This is where we have to consider using interactive mapping of the fields. In D7 it was a little easier because we had fewer fields.

The idea of `$domain_options` is to shortcut the prompts. Ideally, the options / prompts would be:

* Domain Administrators: `DOMAIN_ADMIN_FIELD`
* A keyed list of fields defined by other modules

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1108 @@
...
+    $ids = $this->enumerateDomainEntities($entity_type, $old_domain);
+
+    // @TODO is there a problem loading many, possibly big, entities in one go?
+    $entities = $entity_storage->loadMultiple($ids);
+    $count = 0;
+
+    /** @var \Drupal\Core\Entity\EntityInterface $entity */
+    foreach($entities as $entity) {
+      $changed = FALSE;
+      $value = $entity->get(DOMAIN_ACCESS_FIELD)->value();
```
agentrickard: Written well, this field would be an argument so this routine could run for all of our entity fields.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1108 @@
...
+    }
+  }
+
+  /**
+   * Reassign entities of the supplied type to the $policy domain.
+   *
+   * @return string[]
+   *   List of entities supporting domain access.
+   */
+  protected function findDomainEnabledEntities() {
```
agentrickard: This is a really smart abstraction 👍

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1108 @@
...
+    foreach ($domains as $domain) {
+      $row = [];
+      foreach ($keys as $key) {
+        $v = $domain->get($key);
+        switch($key) {
+          case 'status':
+            if (($options['inactive'] && $v) || ($options['active'] && !$v)) {
+              continue 3; // switch, for, for
+            }
+            $v = ($v) ? dt('Active') : dt('Inactive');
```
agentrickard: I find this shorthand `$v = ($v)` hard to parse. Id rather use `!empty($v) ? dt('Active') : dt('Inactive');`

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1108 @@
...
+   * @usage drush domain-test
+   * @usage drush domain-test example.com
+   * @usage drush domain-test 1
+   *
+   * @command domain:test
+   * @aliases domain-test
+   * 
+   * @throws \Drupal\domain\Commands\DomainCommandException
+   */
+  public function test($domain_id, array $options = ['base_path' => null]) {
```
agentrickard, line 885: See `checkHTTPResponse` above, which should probably contain most of this checking code.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1108 @@
...
+    $domain_storage = $this->getStorage();
+
+    // TODO: This won't work in a subdirectory without a parameter.
```
agentrickard, line 888: Not sure this is still true?

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1108 @@
...
+   *   The machine-readable name to use for the domain.
+   * @usage drush domain-machine-name example.com foo
+   * @usage drush domain-machine-name 1 foo
+   *
+   * @command domain:machine-name
+   * @aliases domain-machine-name
+   *
+   * @throws \Drupal\domain\Commands\DomainCommandException
+   */
+  public function machineName($domain_id, $machine_name) {
```
agentrickard: This was a super-admin tool in D7 when we didn't use machine name for data storage, we just used numeric id. The field data now uses machine_name as the value, which makes this command very volatile. 

If we change a machine name, we have to change any field storage reference to the old machine name.

`enumerateDomainEntities` can help with that check.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1108 @@
...
+    foreach($field_map as $type => $fields) {
+      if (array_key_exists(DOMAIN_ACCESS_FIELD, $fields)) {
+        $entities[] = $type;
+      }
+    }
+    return $entities;
+  }
+
+  /**
+   * Enumerate entities of the supplied type and domain to the $policy domain.
```
agentrickard: Probably also needs a $field argument.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1108 @@
...
+    }
+    else {
+      $this->logger()->info(dt('Nothing to do?'));
+      return;
+    }
+    // Create list of domain (entity) ids to delete.
+    $delete_domain_ids = array_map(function($v){return $v->id;}, $domains);
+    //endregion
+
+    //region Get content disposition from configuration and validate.
```
agentrickard: These regions likely need to be calls to the registry we discussed, where each field acts in it's own method. Those methods like go in `DomainElementManager` and are modeled on `getFieldValues`

domain/src/Commands/DomainCommands.php
rivimey: Done

domain/src/Commands/DomainCommands.php
rivimey: cool

domain/src/Commands/DomainCommands.php
rivimey: Refactored... was in some of my newest code.

domain/src/Commands/DomainCommands.php
rivimey: Going to have to think further about this area.

domain/src/Commands/DomainCommands.php
rivimey: Done

domain/src/Commands/DomainCommands.php
rivimey: Sorted

domain/src/Commands/DomainCommands.php
rivimey: As discussed, yes. Improved sentence.

domain/src/Commands/DomainCommands.php
rivimey: Sorted.

domain/src/Commands/DomainCommands.php
rivimey: Done.

domain/src/Commands/DomainCommands.php
rivimey: Haven't done anything of note for this fn, just the basics. Need to look into it further.

domain/src/Commands/DomainCommands.php
rivimey: ok, might park that one and disable this fn. Sounds like a reassign all over again...

domain/src/Commands/DomainCommands.php
rivimey: I'll have a think about this too.

domain/src/Commands/DomainCommands.php
rivimey: Done.

agentrickard:
I think I fixed the intermittent ResponseTest failures with commit 

https://github.com/agentrickard/domain/pull/434/commits/15275090e133364906f9e27d6d27e0721753402e

So you might want to cherry-pick that over.

PR #432, opened 2018-06-18
Condition false now returns a form item we do not want.

agentrickard:
See https://www.drupal.org/project/domain/issues/2979656


agentrickard:
Intermitted test fail is unrelated.

Issue #427, opened 2018-06-11
"login_paths" install/default setting includes \r\n literally, which doesn't work.


rivimey:
The domain settings form has an install/default value for login_paths of "/user/login\r\n/user/password", which is treated literally, that is \r is not \013 and \n is not \011. The value that counts appears to be the one in domain.settings.yml. Although there is _also_ a value listed in DomainSettingsForm as the second parameter to $config->get():

      '#default_value' => $config->get('login_paths', "/user/login\r\n/user/password"),

... the get() function doesn't have a second parameter so the 'default' value is unused.

[Tests show that at the place this is used, class DomainAccessCheck, function `matchPath` doesn't do the right thing:
    return !$this->pathMatcher->matchPath($path, $allowed_paths);
fails when allowed_paths has the install/default value and $path is "/user/login".]

The appropriate fix would seem to be:

 - modify the login_paths yml value to be an array of lines, so it can be set in domain.settings.yml as:

```
    login_paths:
      - '/user/login'
      - '/user/password'
```

 and adjust get() calls appropriately.

- modify the erroneous #default_value config get() to use this or an equivalent value also.
  + Q: can we read the default for the form directly from domain.settings.yml?

agentrickard:
This is a Drupal 7-ism, but it's consistent with what core does.

This is the output of a block configuration that restricts by path:

```
visibility:
  request_path:
    id: request_path
    pages: "/this\r\n/that"
    negate: false
    context_mapping: {  }
```

I think the fix here is to fix the linebreak handling of the form field and to remove the default values, which have been deprecated -- https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Config%21Config.php/function/Config%3A%3Aget/8.5.x

agentrickard:
The suggested change would therefore break the scheme, since:

```
login_paths:
  - '/user/login'
  - '/user/password'
```

Is an array, not text.

```
Testing Drupal\Tests\domain\Functional\DomainInactiveTest E 1 / 1 (100%)

Time: 37.49 seconds, Memory: 6.00MB

There was 1 error:

1) Drupal\Tests\domain\Functional\DomainInactiveTest::testInactiveDomain
InvalidArgumentException: The configuration property login_paths.0 doesn't exist.
```

I don't think core has this issue because no blocks ship with default values. I think we want to handle this in the form element display.

agentrickard:
I cannot replicated the reported failure here.

agentrickard:
[...]

PR #433, opened 2018-06-18
Better handling of default settings for inactive domain visibility.

agentrickard:
See the report in #427


rivimey:
Looks good to me. The Travis fail is because php7.2 was/is not available so unrelated.

PR #434, opened 2018-06-27
Domain config ui

agentrickard:
The new branch for Domain Config UI.


davidgrayston:
I started to move this into a separate [domain_admin_ui](https://github.com/davidgrayston/domain_admin_ui) module last year, which was abandoned in the end.

Something we may need to include is a cache context for the current selected domain, something like this: https://github.com/davidgrayston/domain_admin_ui/blob/8.x-1.x/src/Cache/Context/SelectedDomain.php

davidgrayston:
Happy to issue a PR to this branch if you feel this cache context is needed

agentrickard:
I don't think we need the cache context because the core `url.site` context is sufficient. In other places in the code, that's what we use, for instance in https://github.com/agentrickard/domain/blob/8.x-1.x/domain/src/Plugin/Condition/Domain.php#L139

davidgrayston:
Ok, just thought I’d mention as this implementation is switching the config based on a session variable (URL won’t change)

agentrickard:
I hadn't considered that. However, I don't think admin forms are cached except by session, so I don't think we have any issues.

We could write a test that has two accounts edit the same form to see if caching is an issue.

agentrickard:
There are some questions in the TODO here -- https://github.com/agentrickard/domain/pull/434/commits/9865bd758d1f2aa083f7a64be4953672ebad1e7c -- that could use review.

domain_config_ui/src/Config/ConfigFactory.php
```diff
@@ -0,0 +1,196 @@
...
+    $this->domainConfigUIManager = $domain_config_ui_manager;
+  }
+
+  /**
+   * Helper to check if config is allowed to be saved for domain.
+   *
+   * @param string $name
+   *   The configuration name.
+   */
+  protected function isAllowedDomainConfig(string $name) {
```
BramG: This is PHP7 only code, the string type hint should be removed

zerolab: Type hinting (in this form) has been in PHP since version 5

BramG: Most type hints are supported but not 'string' as primitive type. That one is added in php 7.0, see http://php.net/manual/en/migration70.new-features.php

BramG: BTW: I noticed this because I got a hard failure on my php 5.6 setup. So I'm 100% sure ;).

zerolab: Ah, was using PHP 7 for so long, I missed that. Fair comment.

agentrickard: Yeah, we don't type hint strings anywhere else. But oddly this doesn't fail here -- https://travis-ci.org/agentrickard/domain/builds/398266990?utm_source=github_status&utm_medium=notification

BramG:
Module works perfectly btw :)

agentrickard:
It works but has some UI quirks that I am sorting through.

agentrickard:
If you are following this issue, I just added code that lets the admin add new forms via the UI, using an AJAX callback.

screen shot 2018-10-17 at 3 55 24 pm

screen shot 2018-10-17 at 3 56 12 pm

domain_config_ui/src/Form/DeleteForm.php
```diff
@@ -0,0 +1,99 @@
...
+      $url = Url::fromRoute('domain_config_ui.list');
+      return new RedirectResponse($url->toString());
+    }
+
+    $elements = DomainConfigUIController::deriveElements($config_name);
+    $config = \Drupal::configFactory()->get($config_name)->getRawData();
+
+    $form['help'] = [
+      '#type' => 'item',
+      '#title' => SafeMarkup::checkPlain($config_name),
```
Lappro: This method is marked deprecated. As discussed in https://www.drupal.org/node/2549395
From the list of suggested replacements `Html::escape` seems closest to the intended behavior.

```suggestion
      '#title' => Html::escape($config_name),
```
This suggestion also needs an additional use statement: `use Drupal\Component\Utility\Html;` and the SafeMarkup use can be removed with this change.

domain_config_ui/src/Form/SwitchForm.php
```diff
@@ -0,0 +1,198 @@
...
+    // Only allow access to domain administrators.
+    $form['#access'] = $this->canUseDomainConfig();
+    $form = $this->addSwitchFields($form, $form_state);
+    return $form;
+  }
+
+  /**
+   * Determines is a user may access the domain-sensitive form.
+   */
+  public function canUseDomainConfig() {
```
paulvandenburg, line 83: I think it would be better if we convert this check to testing if the user is assigned to more than 1 domain.
As the access to some specific settings form should be determined by that other settings form. If the user does have access it should be allowed to modify the settings for all domains he/she has been assigned to.

As a result I think the `field_domain_admin` can be removed. Simplifying the working/configuration of the domain config ui.

domain_config_ui/src/Form/DeleteForm.php
paulvandenburg: This method is marked deprecated. As discussed in https://www.drupal.org/node/2549395
From the list of suggested replacements `Html::escape` seems closest to the intended behavior.

```suggestion
      '#title' => Html::escape($config_name),
```
This suggestion also needs an additional use statement: `use Drupal\Component\Utility\Html;` and the SafeMarkup use can be removed with this change.

paulvandenburg (changes requested):

domain_config_ui/src/Form/SwitchForm.php
agentrickard: That's an advanced behavior that is on the roadmap once we get the UI locked down.

leo72:
Hi,
I did try this on fresh Drupal 8.6.3.
When try to enable for bartik theme gettign error:
Symfony\Component\Routing\Exception\MissingMandatoryParametersException: Some mandatory parameters are missing ("theme") to generate a URL for route "system.theme_settings_theme". in Drupal\Core\Routing\UrlGenerator->doGenerate() (line 182 of /core/lib/Drupal/Core/Routing/UrlGenerator.php).

The full trace

Drupal\Core\Routing\UrlGenerator->getInternalPathFromRoute('system.theme_settings_theme', Object, Array, Array) (Line: 293)
Drupal\Core\Routing\UrlGenerator->generateFromRoute('system.theme_settings_theme', Array, Array, 1) (Line: 105)
Drupal\Core\Render\MetadataBubblingUrlGenerator->generateFromRoute('system.theme_settings_theme', Array, Array, ) (Line: 753)
Drupal\Core\Url->toString() (Line: 43)
Drupal\domain_config_ui\Controller\DomainConfigUIController->ajaxOperation('system.theme_settings_theme', 'enable')
call_user_func_array(Array, Array) (Line: 123)
Drupal\Core\EventSubscriber\EarlyRenderingControllerWrapperSubscriber->Drupal\Core\EventSubscriber\{closure}() (Line: 582)
Drupal\Core\Render\Renderer->executeInRenderContext(Object, Object) (Line: 124)
Drupal\Core\EventSubscriber\EarlyRenderingControllerWrapperSubscriber->wrapControllerExecutionInRenderContext(Array, Array) (Line: 97)
Drupal\Core\EventSubscriber\EarlyRenderingControllerWrapperSubscriber->Drupal\Core\EventSubscriber\{closure}() (Line: 151)
Symfony\Component\HttpKernel\HttpKernel->handleRaw(Object, 1) (Line: 68)
Symfony\Component\HttpKernel\HttpKernel->handle(Object, 1, 1) (Line: 57)
Drupal\Core\StackMiddleware\Session->handle(Object, 1, 1) (Line: 47)
Drupal\Core\StackMiddleware\KernelPreHandle->handle(Object, 1, 1) (Line: 99)
Drupal\page_cache\StackMiddleware\PageCache->pass(Object, 1, 1) (Line: 78)
Drupal\page_cache\StackMiddleware\PageCache->handle(Object, 1, 1) (Line: 47)
Drupal\Core\StackMiddleware\ReverseProxyMiddleware->handle(Object, 1, 1) (Line: 52)
Drupal\Core\StackMiddleware\NegotiationMiddleware->handle(Object, 1, 1) (Line: 23)
Stack\StackedHttpKernel->handle(Object, 1, 1) (Line: 669)
Drupal\Core\DrupalKernel->handle(Object) (Line: 19)

leo72:
Get around the issue above: use wildcard in ENABLED CONFIGURATION FORMS. /admin/appearance/settings/* in my case.
Continue testing...

leo72:
Now the issue is theme color setting (palette for example) doesn't work as it should. I have default domain with palette ONE and second domain with same theme, but palette TWO. Clearing cache makes one of this two palettes applied to both domains :-(

agentrickard:
The issue with theme colors doesn't surprise me at all. The problem with form-based handling is that _some_ forms require special handling which we cannot account for without custom code.

Similar cache problems surface if you try to save field settings per domain, and there is almost nothing that we can do about it.

agentrickard:
See #453 for one approach to the color settings issue, but I don't plan to look at that until this branch is done.

leo72:
Thanks Ken for reply.
I will have a look into #453. How can I help to complete this branch?

agentrickard:
What's missing from this branch are tests and documentation. I think the functionality -- especially the AJAX bit to register forms -- is right where I'd like it.

I got a bit caught up in Drush 9 testing...

leo72:
I ended up by  altering **AssetResolve**

```
class MyModuleServiceProvider extends ServiceProviderBase
{
  public function alter(ContainerBuilder $container) {

    $definition = $container->getDefinition('asset.resolver');
    $definition->setClass('Drupal\my_module\ServiceProvider\MyModuleAssetResolver');
  }
}
```

and in MyModuleAssetResolver overwrite getCssAssets method and add $domain_id to $cid

```
class MyModuleAssetResolver extends AssetResolver
{
  /**
   * {@inheritdoc}
   */
  public function getCssAssets(AttachedAssetsInterface $assets, $optimize) {
    $theme_info = $this->themeManager->getActiveTheme();
    // Add the theme name to the cache key since themes may implement
    // hook_library_info_alter().
    $domain_id = \Drupal::service('domain.negotiator')->getActiveDomain()->id();
    $libraries_to_load = $this->getLibrariesToLoad($assets);
    $cid = 'css:' . $domain_id . ':' . $theme_info->getName() . ':' . Crypt::hashBase64(serialize($libraries_to_load)) . (int) $optimize;
    if ($cached = $this->cache->get($cid)) {
      return $cached->data;
    }
.....
// Rest code of the method here
```

It resolves the issue with color module. What do you think?

agentrickard:
This is in line with the type of changes we have to make for DomainConfigServiceProvider and similar overrides of core services.

Since the Asset Resolver is specific to themes, it makes sense to make that part of Domain Theme -- which I haven't started working on.

gaydabura:
some updates:
- Interface added
- rebased against current version
https://www.drupal.org/project/domain/issues/3004243#comment-12945789

agentrickard:
If you've made any changes, what I really need is a patch or PR against this branch.

gaydabura:
> If you've made any changes, what I really need is a patch or PR against this branch.

[DomainConfigUIManagerInterface_added.txt](https://github.com/agentrickard/domain/files/2803646/DomainConfigUIManagerInterface_added.txt)

agentrickard:
Sweet. Thank you.

agentrickard:
Merged.

agentrickard:
You can pull an updated patch from https://patch-diff.githubusercontent.com/raw/agentrickard/domain/pull/434.patch

JTubex:
This branch no longer applies on the latest release of domain module (1.0.0-alpha16).

agentrickard:
Branches patch against -dev on drupal.org and -dev is stable.

PR #436, opened 2018-07-13
Ruth/drush9 conversion

rivimey:
This is at a place where I want to get serious feedback on it. I commented out the machine name rename following @agentrickard's note : to do this one properly requires code similar to a reassign.

I think some of the code in the commands should be bumped into the main module API, such as identifying domain-enabled entity types, listing entities within a domain, and possibly even part of the reassign code to do with bulk changes.

I tried to get some drush 9 tests done, but so far have timed out trying to get the test framework to work, so the command tests themselves are not tested! Any help with this appreciated; I've already spent hours looking at the code and for the very sparse docs on this area.

Manual testing has been done: I can generate new domains, delete them, list them in various ways, change parameters (e.g. http/s) and get some stats on their use.

There may be some improvements in the area of returned data and what is logged; I think it would be nice to split the reassign entity content out from delete, too.


agentrickard:
I'll try to set aside some time this weekend. 

I'm not too worried about drush tests at the moment.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+ * See these files for an example of injecting Drupal services:
+ *   - http://cgit.drupalcode.org/devel/tree/src/Commands/DevelCommands.php
+ *   - http://cgit.drupalcode.org/devel/tree/drush.services.yml
+ */
+class DomainCommands extends DrushCommands {
+
+  /** @var \Drupal\domain\DomainStorageInterface $domain_storage */
+  protected $domain_storage = NULL;
+
+  protected $entity_field_map = NULL;
```
agentrickard, line 40: Needs documentation

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+
+  protected $is_dry_run = FALSE;
```
agentrickard, line 48: Needs documentation.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+   * @param DomainInterface[] $domains
+   *   List of domains.
+   * @param int[] $exclude
+   *   List of domain names to exclude from the list.
+   * @param DomainInterface[] $initial
+   *   Initial value of list that will be returned.
+   *
+   * @return array
+   */
+  protected function filterDomains($domains, $exclude, $initial = []) {
```
agentrickard: Type hint `array $domains, array $exclude, array $initial`?

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+    foreach ($domains as $one) {
```
agentrickard: Nitpick: $domain instead of $one here.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+   * Check the domain response.
+   *
+   * @param \Drupal\domain\DomainInterface $domain
+   *
+   * @return string
+   */
+  protected function checkHTTPResponse(DomainInterface $domain, $validate_url = FALSE) {
+    // Ensure the url is rebuilt.
+    if ($validate_url) {
+      $this->validateDomain($domain->getHostname());
```
agentrickard: We want to use `DomainValidator::checkResponse()` here, not the full validation routine in `DomainValidator::validate()`

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+  /**
+   * Deletes a domain record.
+   *
+   * @param DomainInterface[] $domains
+   *   The domain_id to delete. Pass 'all' to delete all records.
+   *
+   * @throws \Drupal\domain\Commands\DomainCommandException
+   * @throws \UnexpectedValueException
+   */
+  protected function deleteDomain($domains) {
```
agentrickard: Type hint array.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+    return $entities;
+  }
+
+  /**
+   * Determine whether or not a given entity is domain-enabled.
+   *
+   * @return string[]
+   *   List of entities supporting domain access.
+   */
+  protected function entityHasDomainField(string $entity_type, string $field = DOMAIN_ACCESS_FIELD) {
```
agentrickard: This might become a public method on the DomainElementManager.

I can see a use-case for that.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+      $this->entity_field_map = $entity_manager->getFieldMap();
+    }
+  }
+
+  /**
+   * Enumerate entities of the supplied type and domain.
+   *
+   * @param string $entity_type
+   *   The entity type name, e.g. 'node'
+   * @param string $domain
```
agentrickard: $domain should always refer to a full entity, so we should rename this to $id, $machine_name, or $domain_id

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+  protected function ensureEntityFieldMap() {
+    // Try to avoid repeated calls to getFieldMap() assuming it's expensive.
+    if (empty($this->entity_field_map)) {
+      $entity_manager         = \Drupal::entityManager();
+      $this->entity_field_map = $entity_manager->getFieldMap();
+    }
+  }
+
+  /**
+   * Enumerate entities of the supplied type and domain.
```
agentrickard: I really don't understand what `enumerate` means in this context. The method name needs some clarification.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+   *   the old_domain, this domain replaces it.
+   * @param array $ids
+   *   List of entity IDs for the selected domain and all of type $entity_type.
+   *
+   * @return int
+   *
+   * @throws \Drupal\Component\Plugin\Exception\PluginException
+   * @throws \Drupal\Core\Entity\EntityStorageException
+   */
+  protected function reassignEntities(string $entity_type, string $field, DomainInterface $old_domain, DomainInterface $new_domain, $ids) {
```
agentrickard: Type hint `array $ids`

agentrickard: This method also might move to the DomainElementManager, which needs a slight docs update to say `Handles field management for domain entity references.`

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+    switch($policy) {
+      case 'default':
+        $new_domain = $domain_storage->loadDefaultDomain();
+        break;
+
+      case 'prompt':
+      case 'ignore':
+        return NULL;
+
+      default:
```
agentrickard, line 437: This gets confusing since we have a `case 'default'` above, suggesting that we need to renamed the policy option `default` to something else?

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+        $new_domain = $domain_storage->load($policy);
+        break;
+    }
+    return $new_domain;
+  }
+
+  /**
+   * Reassign entities of the supplied type to the $policy domain.
+   *
+   *  $options = [
```
agentrickard: Is this a `@param`?

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+   *  ];
+   *
+   * @param DomainInterface[] $domains
+   *   List of the domains to reassign content away from.
+   *
+   * @throws \Drupal\domain\Commands\DomainCommandException
+   */
+  protected function reassignLinkedEntities($domains, array $options = ['chatty' => null, 'policy' => null]) {
+    $entity_typenames = $this->findDomainEnabledEntities();
+    // DOMAIN_ADMIN_FIELD too??
```
agentrickard, line 461: DOMAIN_ADMIN_FIELD would be a new operation.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+        // For each domain being reassigned from...
+        foreach ($domains as $domain) {
+
+          // And, For each domain field ...
+          foreach ($field_names as $field) {
+            $ids = $this->enumerateDomainEntities($name, $domain->id(), $field);
+            if (!empty($ids)) {
+
+              try {
+                if ($options['chatty']) {
```
agentrickard, line 483: Isn't `verbose` the most standard option name?

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+        case 'count_active':
+          $v = count($active_domains);
+          break;
+        case 'default_id':
+          $v = $domain_storage->loadDefaultDomain()->getDomainId();
+          break;
+        case 'default_host':
+          $v = $domain_storage->loadDefaultDomain()->getHostname();
+          break;
+        case 'scheme':
```
agentrickard, line 686: Technically, scheme can vary on a per-domain basis, so perhaps we need to list count by scheme? Or this should be `default_scheme`.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+    }
+
+    // Display which entities are enabled for domain access by checking for the fields.
+    $entity_manager = \Drupal::entityManager();
+    $field_map = $entity_manager->getFieldMap();
+    $domain_entities = [];
+    $domain_affiliate_entities = [];
+    foreach($field_map as $type => $fields) {
+      if (array_key_exists(DOMAIN_ACCESS_FIELD, $fields)) {
+        $domain_entities[] = '+'.$type;
```
agentrickard: formatting of concat

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+   * @return string
+   *   The entity id of the created domain.
+   *
+   * @throws \Drupal\domain\Commands\DomainCommandException
+   */
+  public function add($hostname, $name, array $options = ['inactive' => null, 'https' => null, 'weight' => null, 'is_default' => null, 'validate' => null]) {
+    $domain_storage = $this->getStorage();
+    $records_count = $domain_storage->getQuery()->count()->execute();
+    $start_weight = $records_count + 1;
+    $hostname = mb_strtolower($hostname);
```
agentrickard: Normally, we throw an error in the validation step rather than silently correcting the lowercase rule.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+   *   their content and users to the default domain, example.org.
+   *
+   * @option chatty
+   *   Document each step as it is performed.
+   * @option dryrun
+   *   Do not do anything, but explain what would be done. Implies --chatty.
+   * @option content-as
+   *   Values "prompt", "ignore", "default", <name>, Reassign content
+   *   associated with the the domain being deleted to the default domain, to
+   *   a specified domain, or leave the content alone (&hence inaccessible in
```
agentrickard: &hence / and hence

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+   *   the normal way). The default value is 'prompt': ask which domain to use.
+   * @option users-as
+   *   Values "prompt", "ignore", "default", <name>, Reassign user accounts
+   *   associated with the the domain being deleted to the default domain,
+   *   to the domain whose machine name is <name>, or leave the user accounts
+   *   alone (&hence inaccessible in the normal way). The default value is
```
agentrickard: &hence / and hence

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+   *   'prompt': ask which domain to use.
+   *
+   * @command domain:delete
+   * @aliases domain-delete
+   * 
+   * @throws \Drupal\domain\Commands\DomainCommandException
+   */
+  public function delete($domain_id, $options = ['content-as' => null, 'users-as' => null, 'dryrun' => null, 'chatty' => null ]) {
```
agentrickard, line 882: I've not seen cases of pre-populating arrays like this in other code. I don't think the type hint is the proper place to do so.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+    $policy_content = 'prompt';
+    $policy_users = 'prompt';
+
+    $this->is_dry_run = (bool)$options['dryrun'];
```
agentrickard: `(bool) $options`

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+
+    //region Get current domain list and perform sanity checks.
```
agentrickard: We don't use // region annotations.

I also prefer not to use the term `sanity check` and prefer `validations` or similar.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+    //endregion
+
+    //region Determine which domains to be deleted.
+    if ($domain_id == 'all') {
+      $domains = $all_domains;
+      if (empty($domains)) {
+        $this->logger()->info(dt('There are no domains to delete.'));
+        return;
+      }
+      $really = $this->io()->confirm(dt('This action cannot be undone. Continue?:'), FALSE);
```
agentrickard, line 907: What is `$this->io()`?

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+   * @throws \Drupal\domain\Commands\DomainCommandException
+   */
+  public function defaultDomain($domain_id, array $options = ['validate' => null]) {
+    $domain_storage = $this->getStorage();
+    // Resolve the domain.
+    if (!empty($domain_id) && $domain = $this->getDomainFromArgument($domain_id)) {
+      $validate = ($options['validate']) ? 1 : 0;
+      $domain->addProperty('validate_url', $validate);
+      if ($error = $this->checkHTTPResponse($domain)) {
+        throw new DomainCommandException(dt('Unable to verify domain !dom: !error',
```
agentrickard: `!domain` not `!dom`

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+          ['!domain' => $domain->getHostname()]));
+      }
+      if ($domain->status()) {
+        return 'enabled';
+      }
+      else {
+        return 'disabled';
+      }
+    }
+    return 'unknown domain';
```
agentrickard, line 1138: Needs `dt()`? Needs capitalization?

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+  public function renameDomain($domain_id, $name) {
+    // Resolve the domain.
+    if ($domain = $this->getDomainFromArgument($domain_id)) {
+      $domain->saveProperty('name', $name);
+      return $domain->name();
+    }
+    return 'unknown domain';
+  }
+
+  /**
```
agentrickard, line 1204: I think we should just delete this command.

domain/src/Commands/DomainCommands.php
```diff
@@ -0,0 +1,1288 @@
...
+   *    config, etc.
+   *
+   * @usage drush domain-scheme example.com --https
+   * @usage drush domain-scheme 681 https
+   *
+   * @option http
+   *   Set the domain access scheme to http (no SSL).
+   * @option https
+   *   Set the domain access scheme to https (with SSL).
+   *
```
agentrickard, line 1221: There is a third option now: `variable`, which `Sets the domain's URL scheme to inherit from the request.`

agentrickard (changes requested):
Lots of great work here! I've gone through and noted places where I think we can abstract code into the main module -- largely into DomainElementManager.

We should also look into the idea of registering field reassignment via module hook/plugin, though that can wait until this refactor is done.

domain/src/Commands/DomainCommands.php
rivimey: Ok, I have adapted that code. not sure the new version is correct as I realized this code was expected to return something, and it wasn't. Now returning TRUE if asked to validate and the response code is 2xx, or not validating.

domain/src/Commands/DomainCommands.php
rivimey: Done

domain/src/Commands/DomainCommands.php
rivimey: Done

domain/src/Commands/DomainCommands.php
rivimey: Done.

domain/src/Commands/DomainCommands.php
rivimey: Done.

domain/src/Commands/DomainCommands.php
rivimey: It is part of the new drush 9 command infrastructure, which replaces the old drush_print, drush_choice et al. The call returns a configured IO object which is a subclass of SymfonyStyle, and is provided by the DrushCommands class.

domain/src/Commands/DomainCommands.php
rivimey: Done

domain/src/Commands/DomainCommands.php
rivimey: I'm not sure. Drush now more than before encourages subcommands to return data so it can be processed by the drush output system, so here I am returning a string which will be printed, possibly as json, ...

If this is the intent, then dt() is not really wanted, because we want predictable returned value. However for human invocations then dt() would be appropriate.

domain/src/Commands/DomainCommands.php
rivimey: Sorry, got confused. Do you mean delete the "drush domain:disable" and "drush domain:enable" commands, or delete the "drush domain:machine-name" command which I commented out?

agentrickard: machine name

agentrickard: machine name

agentrickard: machine name

agentrickard: machine name

domain/src/Commands/DomainCommands.php
agentrickard: This is a drush change that I'm not aware of, so this seems fine. Any idea how this is reported? That would answer the capitalization question.

domain/src/Commands/DomainCommands.php
rivimey: I have added the third option and tidied up the code somewhat as well; the old "drush scheme http" is now "drush domain:scheme --set=http" as I feel this is a lot clearer what is happening. I have also modified and updated the code relating to scheme in the domain:add command.

rivimey: A thought: should there be a standard function somewhere that answers the question "Is this string a valid scheme name"?

rivimey:
The drush command testing code is preventing the test suite from completing, because it requires drush to be present in the test environment (obvs) and it is not currently there. I'm not sure how to request this happens, so all told I think it might be better to split the file off into another branch to be completed at another time... do you agree?

domain/src/Commands/DomainCommands.php
agentrickard: I don't think so. We only support `http` and `https` so those are hardcoded in most places. Even here, the input is numeric and we convert to a fixed string.

agentrickard:
Yes, we can split the new test file off to a new branch.

Drush is installed by the Travis file, but it's drush 8. So this might be simple to fix:

```
  # Install Drush.
  - composer global require --no-interaction drush/drush:8.*
```

However, all the drush commands in the `.travis.yml` file may need updating as a result.

rivimey:
I've been off sick for some time now (and still not well) and so unable to access my dev environment. However I have "disabled" the two drush test files which were causing problems, and rebased to the current head. Could you let me know where we are with acceptance?

domain/composer.json
```diff
@@ -0,0 +1,21 @@
+{
+    "name": "org/domain",
```
agentrickard: `org`? Should that be `drupal`?

domain/composer.json
```diff
@@ -0,0 +1,21 @@
...
+    "description": "This extension provides new commands for Drush.",
+    "type": "drupal-drush",
+    "authors": [
```
agentrickard, line 6: We should fill this in or remove it.

domain/composer.json
```diff
@@ -0,0 +1,21 @@
...
+            "email": "author@example.com"
+        }
+    ],
+    "require": {
+        "php": ">=5.6.0"
+    },
+    "extra": {
+        "drush": {
+            "services": {
+                "drush.services.yml": "^9"
```
agentrickard, line 20: Just so I understand -- this means that to support Drush 8 we need to keep the domain.drush.inc file?

domain/src/DomainValidator.php
```diff
@@ -133,7 +133,7 @@ public function checkResponse(DomainInterface $domain) {
     }
     // We cannot know which Guzzle Exception class will be returned; be generic.
     catch (RequestException $e) {
-      watchdog_exception('domain', $e);
+      //watchdog_exception('domain', $e);
```
agentrickard, line 136: Do you know why this was throwing errors?

domain/tests/src/Functional/DomainCommandTestBase.txt
```diff
@@ -0,0 +1,60 @@
...
+  protected $profile = 'standard';
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp() {
+    parent::setUp();
+
+    if (UNISH_DRUPAL_MAJOR_VERSION < 8) {
+      $this->markTestSkipped('Migrate manifest is for D8');
```
agentrickard: Error text update.

domain/tests/src/Functional/DomainDrushCommandsTest.txt
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * Show some info about the domain module environment.
+   */
+  public function testInfoDomains() {
+    if ($this->isWindows()) {
+      $this->markTestSkipped('Site-set not currently available on Windows.');
+    }
+    $this->setUpDrupal(1, true);
+
+    // Assure that a pending post-update is reported.
```
agentrickard: This whole section doesn't mean anything to me, as the comment doesn't match the code.

It's also not clear how 'domain_access_entities' is getting set.

domain/tests/src/Functional/DomainDrushCommandsTest.txt
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * List the domains present on the system and some of their properties.
+   */
+  public function testListDomains() {
+    if ($this->isWindows()) {
+      $this->markTestSkipped('Site-set not currently available on Windows.');
+    }
+    $this->setUpDrupal(1, true);
+
+    // Assure that a pending post-update is reported.
```
agentrickard: Same here. just reading it, I suspect that `getOutputFromJSON` is returning a boolean, which is why the test passes.

domain/tests/src/Functional/DomainDrushCommandsTest.txt
```diff
@@ -0,0 +1,105 @@
...
+    $this->drush('domain:add', [ 'my.domain.example.com', 'my_domain_example_com'], ['format' => 'json']);
+    $out = $this->getOutputFromJSON('devel-post-null_op');
+    $this->assertEquals('1', $out->domain_access_entities);
+  }
+
+  /**
+   * List the domains present on the system and some of their properties.
```
agentrickard: Incorrect documentation.

agentrickard (commented):
I love the progress here, and especially the tests. I've marked places where things are unclear to me.

agentrickard:
I'm also about to go on vacation for two weeks, away from a computer.

domain/composer.json
rivimey: My understanding is that drush 8 no longer works with Drupal 8, and the reason it might seem to is that drush 8 will actively look for an instance of drush 9 within the drupal 8 installation, and use that.

The actual requirement here is the 'extra' section including 'drush':
https://docs.drush.org/en/master/commands/#specifying-the-services-file

I have copied the relevant content from the domain/composer.json file into this file: domain/domain/composer.json, which hopefully addresses some of these issues.
It seems possible to me that the former file should not declare itself to be a drupal-module?

domain/tests/src/Functional/DomainCommandTestBase.txt
rivimey: Sorry, I'm not sure what you mean here.

domain/src/DomainValidator.php
rivimey: I'm sorry I don't recall exactly at this point, but I think it was for dns lookup error. I believe I changed this because returning an exception from a validate seemed to me to be an odd thing to do when we are explicitly returning an error code as well.

domain/tests/src/Functional/DomainDrushCommandsTest.txt
rivimey: Now I have looked more carefully, I think the tests are showing an excess of cut-n-pasting. A lot of it is currently rubbish, so I will delete the files. I don't think they are adding anything useful.

domain/src/Commands/DomainCommands.php
rivimey: My intention is to return a list of the entity instances for a given condition. So for 'node' type then return a list of node ids for the supplied domain.
I am not clear where the issue lies? Could you help by saying what you thought it might mean?

domain/src/Commands/DomainCommands.php
rivimey: I have updated the comments around here: hopefully this resolves it.

domain/src/Commands/DomainCommands.php
rivimey: It should be :-)

domain/src/Commands/DomainCommands.php
rivimey: Not really understanding. Should I remove the comment?

domain/src/Commands/DomainCommands.php
rivimey: Sadly drush itself uses verbose, so although it is available to be used by commands you end up seeting lots of drush internal verbose stuff if you use it.

domain/src/Commands/DomainCommands.php
rivimey: Not sure I can deal with this without the dev environment.

domain/src/Commands/DomainCommands.php
rivimey: Replaced with a validation step.

domain/src/Commands/DomainCommands.php
rivimey: I hadn't either, but it seems to be the "standard" for drush, so I was following that example. Previously I have set up a default values array and array_merged() together.

agentrickard:
For some reason, I can't see a lot of the conversation, probably because the files were removed.

I liked having the test files. They gave us a roadmap. We should replace the copy/paste with TODO items.

agentrickard:
Moved to #465

PR #442, opened 2018-09-14
Canonical url

agentrickard:
Fixes https://www.drupal.org/project/domain/issues/2992854


agentrickard:
Injecting the domain access manager into the service here creates a dependency that we may not want.

agentrickard:
Probably needs to use FieldedEntityInterface. See https://www.drupal.org/project/domain/issues/2979581

PR #443, opened 2018-09-28
Issue #3002744 by Shyamala, agentrickard: Domain Config Overrides

agentrickard:
See https://www.drupal.org/node/3002744


zerolab:
The changes look good to me.
And the one test failure has to do with a timeout, so can be ignored (or restarted if needed)

PR #447, opened 2018-10-24
Remove an error inside DomainAliasLoaderInterface

vuchkov:
Remove an error  into DomainAliasLoaderInterface:

`Fatal error: Declaration of Drupal\domain_alias\DomainAliasLoader::loadMultiple($ids = NULL, $reset = false) must be compatible with Drupal\domain_alias\DomainAliasLoaderInterface::loadMultiple(array $ids = NULL, $reset = false) in /var/www/dev91/web/modules/contrib/domain/domain_alias/src/DomainAliasLoader.php on line 14`


vuchkov:
Remove an error  into DomainAliasLoaderInterface

PR #453, opened 2018-10-27
Add domain_assets.module

geek-merlin:
This module varies library info per domain, effectively allowing color settings per domain.

Also needed: Core patch [Color css files are not regenerated when deploying through configuration management \[\#2995825\] \| Drupal\.org](https://www.drupal.org/project/drupal/issues/2995825#comment-12832872) which might love some test fixes and RTBC.

### DIFF
--- a/domain_assets/domain_assets.info.yml
+++ b/domain_assets/domain_assets.info.yml
@@ -1,0 +1,8 @@
+name: Domain Assets
+description: 'Allows domain specific assets, e.g. for color.module.'
+type: module
+package: Domain
+version: VERSION
+core: 8.x
+dependencies:
+  - domain:domain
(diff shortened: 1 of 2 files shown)


geek-merlin:
@agentrickard It looks like the test builds failed on php5.x, and maybe i required php7 somewhere.
Should i fix that given that core requires 7 now or do we follow core?

agentrickard:
I changed things a while back to only test 7.1 and 7.2. This fail is an intermittent, unpredictable fail in `Drupal\Tests\domain\Functional\DomainGetResponseTest ` that we can ignore.

I'm still not entirely sure how I would use this module, though.

geek-merlin:
Yes, reading the phpdoc after some weeks, i realize i'll improve that and move to a README. Some quick notes for this:

The problem: If you use color.module with some domain specific config, or some other libraries (css or js) that vary by domain, then core can not cope with that as LibraryInfo and AssetManager do not understand cache contexts.

This module fixes that by adding a domain-dependent dummy library to every domain, so PageCache and AssetManager both vary their cache key by url.site.

geek-merlin:
Also: If you prefer, i can maintain this in a separate namespace.

PR #454, opened 2018-11-01
[Test-only] domain_alias install fails with existing content

dpagini:
Built a test to demonstrate the failure of `domain_alias` module being installed after a site has content.

I realize the suggestion is present this is a stubborn Drupal cache, but it seems there should be _something_ we could do to help alleviate this issue?

Test for: https://www.drupal.org/project/domain/issues/2976117


agentrickard:
Fascinating. Thank you for writing the test case.

agentrickard:
Moved to #464 for work.

PR #455, opened 2018-11-02
Revert required context.

### LINKED ISSUE #457: Blocks disappears after database update
After a Core (8.6.2) and Domain Access update, all of my blocks have disappeared. Looks like a Domain issue because the bug doesn't appears in several others Drupal 8 that I updated, except those with the Domain Access module.

I've made some tests, and I understand that the bug is caused by the database update 8001 with version 1.0-alpha14.

### Versions

Before the update :

- Drupal : 8.6.1
- Domain : 8.x-1.0-alpha13

After the update :

- Drupal : 8.6.2
- Domain : 8.x-1.0-alpha14

### Explanations

In the back-office, all my blocks are here and I see no changes in their configurations.
The only difference I noticed between a visible block and a hidden one is the `visibility > domain > context_mapping` array.

This array is empty in my hidden blocks...

... and not when my block is visible

dpagini:
This is a suggested revert of b375017ae131db1d638f3a19e5a82371ed22012a.

I am trying to upgrade domain from `alpha13` to `alpha14` and was previously using the domain condition in `page_manager` to select different homepage variants per webpage. It seems when this value changed from FALSE to TRUE, the ability to use Domain within page_manager was removed.

I couldn't find any background other than the commit message on why this was removed, so I was hoping to post this suggestion and see if I could get any more background here? Maybe we could fix this some other way?

_Edit:_ Closes #457

### DIFF
--- a/domain/src/Plugin/Condition/Domain.php
+++ b/domain/src/Plugin/Condition/Domain.php
@@ -15,7 +15,7 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  *   id = "domain",
  *   label = @Translation("Domain"),
  *   context = {
- *     "entity:domain" = @ContextDefinition("entity:domain", label = @Translation("Domain"), required = TRUE)
+ *     "entity:domain" = @ContextDefinition("entity:domain", label = @Translation("Domain"), required = FALSE)
  *   }
  * )
  */


dpagini:
I am assuming the CI failures here are unrelated to this change, but please advise if you don't believe so...?

agentrickard:
So, I am back from vacation and will need to dig into this. The fact that the issue suggests a conflict between core context handling and Page Manager is troubling, though the actual change seems minor.

agentrickard:
The test fails are merely timeouts, and not relevant here.

dpagini:
Thanks @agentrickard. Let me know if/how I can help. I also linked another issue with someone having a very similar problem with block visibility that's worth taking a look at as a 2nd example.

agentrickard:
Thanks. I just have to look to see why I removed that in the first place and if it causes any issues putting it back.

agentrickard:
Yeah, I was trying to remove this default (and redundant) element from the block form. I'll try to find a different way.

screen shot 2018-11-14 at 1 16 29 pm

agentrickard:
Since I can't test off this branch, I had to make a new PR #459 for review.

PR #456, opened 2018-11-06
Issue #3011724 [Domain Source] Set default value in BuildForm()

danny-wouters:
The pull request for issue on https://www.drupal.org/project/domain/issues/3011724

### DIFF
--- a/domain_source/src/Form/DomainSourceSettingsForm.php
+++ b/domain_source/src/Form/DomainSourceSettingsForm.php
@@ -45,7 +45,7 @@ class DomainSourceSettingsForm extends ConfigFormBase {
     $form['exclude_routes'] = [
       '#type' => 'checkboxes',
       '#title' => $this->t('Disable link rewrites for the selected routes.'),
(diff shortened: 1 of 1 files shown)


paulvandenburg:
I don't see why Travis has raised errors. The logs seems clear to me and I can also hardly believe this change could raise an error.

Either way I've confirmed this change to be working.

zerolab:
On of the jobs failed with "The job exceeded the maximum time limit for jobs, and has been terminated.". Could be restarted, but not a big one.

This could probably do with a quick test

ideb:
This error can occur for sites that had domain_source enabled before domain_source.settings was added. The only way to add a test for the failing behavior is to programmatically delete domain_source.settings and open the administration form.

paulvandenburg:
The best test probably would be some code quality test. Which should complain whenever an interface method is called with more arguments than the interface defines.

Because the true cause of this error is a developer accidentally using an interface method incorrectly, which can happen anywhere in the code.

PR #459, opened 2018-11-14
Condition form

agentrickard:
Builds on work from #455 and sets the form element to something readable.

screen shot 2018-11-14 at 1 26 30 pm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,11 +94,11 @@ before_script:
   - until curl -s example.com:8080; do true; done > /dev/null
 
 script:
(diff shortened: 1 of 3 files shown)


dpagini:
Just FYI - I pulled this patch into my project that was having the issues, and can confirm it does fix my problem.

agentrickard:
Thanks.

agentrickard:
See https://www.drupal.org/project/domain/issues/3011055

jcmartinez:
Patching with composer from this PR against 1.0.0-alpha14 didn't apply for me. I had to patch from 455.

dpagini:
Same for me @jcmartinez. I think this ended up with a rogue extra commit on it, so I went back to that other PR.

agentrickard:
That's probably due to the merge conflict noted in 21ffdee

PR #462, opened 2018-11-15
Issue #3010847 by idebr: Enable strictConfigSchema checking in automa…

ideb:
…ted tests for Domain Access and Domain Content


agentrickard:
Sigh. Two of these test failures are timeouts. The other is an intermittent fail in the GetReponseTest that I thought we fixed years ago.

agentrickard:
Rerunning as #479

PR #464, opened 2018-11-26
Issue #2976117 by ZeiP, johnpicozzi, dpagini: Enabling domain_alias fails  on existing site

agentrickard:
Replaces #454


agentrickard:
This test does not fail if the node module is installed.

PR #466, opened 2019-01-12
Failing test for invalid hook cache.

dpagini:
Test to demonstrate issue in https://www.drupal.org/project/domain/issues/3025541

I found there's a problem using domain_config (which injects DomainNegotiator service, which invokes hooks) along with a service that runs early in the drupal bootstrap process, such as a page cache policy service.

I have added two unit tests here. One that runs and operates as expected, and then the same test that fails once `domain_config` gets installed.


agentrickard:
I've mirrored this in #467 so I can test it.

PR #470, opened 2019-02-21
Move all the node access implementation to a service so it can be disabled or modified.

gheydon:
For a client we need to not have the node access integration running but have all the domain access enabled to get the same front end features.

The one downside to this approach is that if you enable the use Null access method which will create no node access records or change how the system operates but Drupal will still recognise there still being a node access module existing, and create a open access record for every node. I concluded that 99.9% of the time if the developer is disabling node access by using the Null access method they will be running another node access module so this will not be an issue.


dpagini:
Couldn't you just do a `hook_node_access_records_alter` and a `hook_node_grants_alter` for the client that needs this behavior?

How would you actually use this... override the service?
```yml
  domain_access.access_method:
    class: Drupal\domain_access\AccessMethod\Null
```

gheydon:
> Couldn't you just do a `hook_node_access_records_alter` and a `hook_node_grants_alter` for the client that needs this behavior?

You could use these hooks, but then if you have multiple node access modules it becomes more complicated to get rid of the domain access records. Also this would have an overhead of both generating of removing these records.

> How would you actually use this... override the service?

Yes you would just override the service in your services.yml

gheydon:
I have done an initial release of [domain advanced](https://www.drupal.org/project/domain_adv/releases/8.x-1.0-alpha1) for Drupal 8.x which needs this PR to run correctly.

agentrickard:
What features, exactly? I would have thought that making Domain Access optional would avoid this issue.

domain_access/src/AccessMethod/NodeAccess.php
```diff
@@ -0,0 +1,149 @@
...
+  public function __construct(DomainAccessManagerInterface $domainAccessManager, DomainNegotiatorInterface $domainNegotiator, EntityTypeManagerInterface $entityTypeManager) {
+    $this->domainAccessManager = $domainAccessManager;
+    $this->domainNegotiator = $domainNegotiator;
+    $this->entityTypeManager = $entityTypeManager;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function Grants(AccountInterface $account, $op) {
```
agentrickard: Incorrect method naming.

domain_access/src/AccessMethod/NodeAccess.php
```diff
@@ -0,0 +1,149 @@
...
+        $grants['domain_id'][] = $id;
+      }
+    }
+    return $grants;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function Records(NodeInterface $node) {
```
agentrickard: Incorrect method naming.

domain_access/src/DomainAccessMethodInterface.php
```diff
@@ -0,0 +1,37 @@
...
+   * Build the grants avaible for passed user.
+   *
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *   Account of the user to generate the accounts for.
+   * @param $op
+   *   The operation to generate the grants for.
+   *
+   * @return array
+   */
+  public function Grants(AccountInterface $account, $op);
```
agentrickard: We should probably call this nodeAccessGrants().

domain_access/src/DomainAccessMethodInterface.php
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * Build node access records.
+   *
+   * @param \Drupal\node\NodeInterface $node
+   *   Node for the records to be created for.
+   *
+   * @return array
+   */
+  public function Records(NodeInterface $node);
```
agentrickard: nodeAccessRecords()?

agentrickard:
Note: Adding hook_node_grants_alter() or hook_node_access_records_alter() does not register a module as a Node Access module.

You generate on node_save() so aren't doing a DB write/delete. The performance impact is minimal.

domain_access/src/AccessMethod/NoRecords.php
```diff
@@ -0,0 +1,30 @@
...
+use Drupal\Core\Session\AccountInterface;
+use Drupal\domain_access\DomainAccessMethodInterface;
+use Drupal\node\NodeInterface;
+
+/**
+ * Class Null.
+ *
+ * @package Drupal\domain_access
+ */
+class NoRecords implements DomainAccessMethodInterface {
```
agentrickard, line 14: I don't think it's our responsibility to provide the NULL service. You'd have to alter the service container to activate this anyway.

Overall, I think the cleaner solution is to answer the question "What front-end features do you need that aren't available without Domain Access?"

domain_access/src/DomainAccessMethodInterface.php
gheydon: I have made this change.

domain_access/src/AccessMethod/NoRecords.php
gheydon: You can create a service provider https://cgit.drupalcode.org/domain_adv/tree/src/DomainAdvServiceProvider.php to make this change on the fly. 

I can remove the NoRecords class and have other modules provide their own implementation.

The Domain access module provides all the content domain assignment code, so you can specify where domains the content is available on. With Domain access advanced I need all the records and everything, I just need the node access records to not be there.

I did investigated the use of the `hook_node_grants_alter()` and `hook_node_access_records_alter()` but unlike the old `hook_domainrecords()` and `hook_domaingrants()` I could just `return array();` and not worry about records from other modules implementing node access.

agentrickard:
I really don't understand how this is possible:

> The Domain access module provides all the content domain assignment code, so you can specify where domains the content is available on. With Domain access advanced I need all the records and everything, I just need the node access records to not be there.

Without the node access records, how is content assigned correctly?

gheydon:
> Without the node access records, how is content assigned correctly?

By themselves this doesn't assign anything. but conversely I don't understand how domain will work without domain access and domain content.

The [Domain access advanced](https://www.drupal.org/project/domain_adv) module then takes alter's the sql directly to which then assigns the content to the correct to the domains.

The Domain access advanced module gets around the issue of having multiple node access modules running at the same time. Since Domain is an all encompassing module in that it affects every node in the system so using a another node access module will cause the content with multiple node access rules which apply to a single node will appear on all domains.

dpagini:
I saw the last comment and wanted to chime in again... I had originally chimed in b/c I was interested in this "topic" b/c I too am using domain module along with 3 other node_access modules. In my opinion, the model Drupal uses for this leaves something to be desired, b/c it was very difficult to "glue" these modules together, but I did it with `hook_node_grants_alter` and `hook_node_access_records_alter` alone.

If this patch were to be accepted to this project... don't you need any of the other modules you are expecting to integrate to ALSO adopt this kind of model? Or no?

zerolab:
Great point @dpagini.
@gheydon have you seen https://drupal.org/project/acb ? It acts as a bridge between the various node access modules without requiring any fundamental changes to any of them. We've been using it in production on large and complex site with Domain Access, OG and TAC

agentrickard:
Gordon-

As it stands, you can use Domain Source (URL rewriting) independent of Domain Access. Domain Content (admin views) does require Domain Access.

Without this patch, you'd need to create your own instance of the Domain node reference form element (which is easy if you don't need it to be sensitive to variable admin privileges). Then you'd need to recreate the views provided by Domain Content.

I'm very hesitant about moving all this to a service, as it makes maintenance (which is already hard) even harder by introducing numerous _external_ points of failure for module users who install derivative modules that I don't control.

I'd greatly prefer to handle all of this in the grants / records alter hooks.

The core database storage for node grants is _already_ a service. So you could swap that out and get more reliable control over how grants are handled.

```
  node.grant_storage:
    class: Drupal\node\NodeGrantDatabaseStorage
    arguments: ['@database', '@module_handler', '@language_manager']
    tags:
      - { name: backend_overridable }
```

gheydon:
I am going to close this pull request as I worked out a new method. Basically turning off `hook_node_grants()` and `hook_node_access_records()` for domain access so they are not called or detected.

Here is the code to turn off the node access integration for Domain access.
```
/**
 * Implements hook_module_implements_alter().
 */
function domain_adv_module_implements_alter(&$implementations, $hook) {
  if ($hook == 'node_grants' || $hook == 'node_access_records') {
    unset($implementations['domain_access']);
  }
}
```

agentrickard:
I totally forgot about that trick. It's brilliantly simple.

agentrickard:
Beware that doesn't unset other hooks accidentally.

PR #473, opened 2019-03-05
Create some handy twig global variables and functions to assist with theming

gheydon:


agentrickard:
Not sure what to make about the failing tests here.

This would also require some documentation around how to use the variables, because I don't understand this code. That documentation could go in README.md

PR #483, opened 2019-07-02
Add option to domain source to stop redirection from aliases

gheydon:
The current default behaviour of Domain source is to direct to the primary domain even if you are on an alias of the current domain that needs to be redirected to.

This caused me huge problems with GraphQL doing a route lookup query

```
query {
  route(path: "/bar") {
    ... on InternalUrl {
      nodeContext {
        __typename
      }
    }
  }
}
```

So in the example of using a primary domain of `foo.com` and the a graphql domain of `api.foo.com` and the Domain source it will try to redirect.

### DIFF
--- a/domain_source/config/install/domain_source.settings.yml
+++ b/domain_source/config/install/domain_source.settings.yml
@@ -1,1 +1,2 @@
 exclude_routes: {}
+dont_redirect_on_aliases: 0
--- a/domain_source/config/schema/domain_source.schema.yml
+++ b/domain_source/config/schema/domain_source.schema.yml
@@ -8,3 +8,6 @@ domain_source.settings:
       sequence:
         type: string
         label: 'Route'
+    dont_redirect_on_aliases:
+      type: boolean
+      label: 'Stop redirecting on aliases'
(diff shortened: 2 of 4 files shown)


agentrickard:
I do not understand the problem as currently explained.

Can you please provide more detail.

gheydon:
Basically if you have a situation when you have a domain and aliases of that domain, Domain source will redirect to the domain to primary domain even if you are on the alias of the domain.

eg. I have a primary domain `https://foo.com` and an alias `https://bar.com` of the primary domain. If you have a node `node/1` with the domain source pointing to the primary domain. This means that on the primary domain the reference to the node will always be absolute `https://foo.com/node/1` and the same on the alias. If you go to `https://bar.com/node/1` you will get redirected to `https://foo.com/node/1` even though they are the same domain, except `foo.com` is the alias.

This is usually not a problem except for the unnecessary redirects but when using GraphQL some queries do not work unless you are on the base url for the domain. Which means you can't do things like use `https://api.foo.com` to do all your queries.

So adding in this additional config option on the Domain source means that you can reduce your redirects and in the case of GraphQL use an `api` domain to make any calls.

agentrickard:
I see. These aren't redirects, per se, (unless the alias is set to redirect to it's parent) but the link HREF is being rewritten.

I wonder if instead of this patch, we should set Domain Source to be alias-aware. It certainly should be, and I would expect that the 'environment' handling would account for that. 

To make that work, the current logic in this PR is close, but not quite right, I think.

Is the essential request here "Allow API calls to ignore Domain Source?"

Even if we fix that, I don't think you can use `api.foo.com` for all queries, because it won't know how to handle content separation unless you're handling that outside the node access system (likely by using a query_alter hook).

PR #486, opened 2019-10-08
Add domain_alias fails when webprofiler module is enabled

JTubex:
https://www.drupal.org/project/domain/issues/3086471


agentrickard:
These failures are not related to the patch/

PR #487, opened 2019-10-11
Port alias refactor

hanoii:
## Better handling of ports with and without default

I found out that the whole alias matching was not working as expected.

Main issue I found is that eventhough an alias like `example.com:*` was defined, it was not properly picked up if accessed by example.com, as/if the port was not found.

Then I found the logic a bit odd and hard to follow, things like `:0` where being added and the foreach was modifying itself so it prompted a bit of a refactor. Basically the port is just found and removed, then the regular patterns are built, and then port alternatives are used. If port is not found on the domain, the actual one is taken from the request and used. If they are `80` or `443`, the exact domain is also added as part of the pattern matching.

## Use setter injection type instead of constructor

I needed to add another dependency and noticed about #477. I also refactored this bit to use a setter injection approach (https://symfony.com/doc/current/service_container/injection_types.html). The main benefit of this is that it becomes independent on what core does (unless it changes the static signature but that's a whole different change). So now you can add internal dependencies and if core decides to add a new one, the module will still work appropriately.

## Drupal.org

https://www.drupal.org/project/domain/issues/3087449

### DIFF
--- a/domain_alias/src/DomainAliasStorage.php
+++ b/domain_alias/src/DomainAliasStorage.php
@@ -2,15 +2,12 @@
 
 namespace Drupal\domain_alias;
 
-use Drupal\Component\Uuid\UuidInterface;
-use Drupal\Core\Cache\MemoryCache\MemoryCacheInterface;
 use Drupal\Core\Config\Entity\ConfigEntityStorage;
 use Drupal\Core\Entity\EntityTypeInterface;
-use Drupal\Core\Language\LanguageManagerInterface;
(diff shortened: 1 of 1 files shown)


agentrickard:
I like this, but we've had issues with injection of the LanguageManager and other services early in the bootstrap process, so those would need testing.

See the test modules in domain_config.

We should also update the `DomainAliasWildcardTest` to see that the changes work. I'd prefer to see the functional change and test change in one patch and the refactor in another, sine this Injection syntax is new to me.

agentrickard:
Note that this also conflicts a little with #486

PR #490, opened 2019-11-22
Currently all tests are failing for 8.8.x branch.

JTubex:

PR #493, opened 2019-12-17
Try new travis configuration

agentrickard:
This is an update of #490, merged with earlier changes.


agentrickard:
This works, though it doesn't seem possible to turn off deprecation warnings, so that will be a follow-up.

PR #507, opened 2020-06-25
Ensure page cache works as expected

agentrickard:
https://www.drupal.org/project/domain/issues/3154402


agentrickard:
This has been folded in to #508

PR #510, opened 2020-08-10
add fields and filters to content overview

malcomio:
See https://www.drupal.org/project/domain/issues/3164346


agentrickard:
The default view deliberately replicates core's UI. These are easy enough to override on a per-site basis that I don't see the value in making this change globally.

PR #512, opened 2020-10-02
Replace deprecations in tests

JTubex:
This PR removes all the deprecated code in the test classes that will be removed in Drupal 10.

Change records: 
- [Overridden test methods require void return type hints](https://www.drupal.org/node/3114724)
- [$modules is now a protected property in tests](https://www.drupal.org/node/2909426)
- [drupalPostForm in functional tests is deprecated](https://www.drupal.org/node/3168858)
- [Simpletest's legacy assertion methods are deprecated](https://www.drupal.org/node/3129738)
- [Calls to WebAssert methods do not allow more arguments than those in the signature](https://www.drupal.org/node/3162537)

### DIFF
--- a/domain_access/src/DomainAccessManager.php
+++ b/domain_access/src/DomainAccessManager.php
@@ -92,7 +92,7 @@ class DomainAccessManager implements DomainAccessManagerInterface {
    * {@inheritdoc}
    */
   public static function getAllValue(FieldableEntityInterface $entity) {
-    return $entity->hasField(DOMAIN_ACCESS_ALL_FIELD) ? $entity->get(DOMAIN_ACCESS_ALL_FIELD)->value : NULL;
+    return $entity->hasField(DOMAIN_ACCESS_ALL_FIELD) ? (bool) $entity->get(DOMAIN_ACCESS_ALL_FIELD)->value : NULL;
   }
 
   /**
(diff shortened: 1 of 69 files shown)


domain_access/src/DomainAccessManager.php
```diff
@@ -92,7 +92,7 @@ public static function getAccessValues(FieldableEntityInterface $entity, $field_
    * {@inheritdoc}
    */
   public static function getAllValue(FieldableEntityInterface $entity) {
-    return $entity->hasField(DOMAIN_ACCESS_ALL_FIELD) ? $entity->get(DOMAIN_ACCESS_ALL_FIELD)->value : NULL;
+    return $entity->hasField(DOMAIN_ACCESS_ALL_FIELD) ? (bool) $entity->get(DOMAIN_ACCESS_ALL_FIELD)->value : NULL;
```
JTubex, line 95: This is the only non-test related change. It seems Drupal 9 added a strict check to assertTrue/assertFalse.

```  /**
   * Get the all affiliates field values from an entity.
   *
   * @param \Drupal\Core\Entity\FieldableEntityInterface $entity
   *   The entity to retrieve field data from.
   *
   * @return bool
   *   Returns TRUE if the entity is sent to all affiliates.
   */
  public static function getAllValue(FieldableEntityInterface $entity);
```

The return value is documented as a boolean but it seems this method is returning integers.

agentrickard:
I have not been tracking issues here, so the https://www.drupal.org/project/domain/issues/3183018 commit broke this PR.

agentrickard:
We also need to know which version(s) of Drupal are supported by these changes. The removal of drupalPostForm was compatible, I believe with 8.8 and higher, which is fine.

I think we're supporting 8.9 and higher right now...

